PR #3, opened 2018-07-25
removes the static flag form the storybook command

theinterned:
We don't need a public folder for static assets in the sandbox as assets can be co-located with stories


nikola-nulogy (approved):
Looks good.

PR #5, opened 2018-07-25
renames sandboxes to sandbox

theinterned:
This is just a quick one to rename `sandboxes/` to `sandbox/`. 

# Why?

Because singular names are better than plural and really we're all playing in the same sandbox right?

(_Also I added restrictions to block merges to master and I wanted to test it_)/

# What I did

1. renamed the directory `sandboxes/` to `sandbox/`
1. updated the package name to `@nulogy/sandbox`
1. updated other occurrences in code.
1. added example stories to use as a reference / starter.

# To test

run `$ yarn hack` ... the storybook should build.


nikola-nulogy (commented):
It works fine. The only thing is that I am still seeing "sandboxes" folder. Is that expected?

theinterned:
@nikola-nulogy I just pulled down afresh clone of the repo and verified that the `sandboxes/` folder is renamed to `sandbox/`. It could be that you have uncommitted changes in the `sandboxes/` folder?

nikola-nulogy (approved):

PR #11, opened 2018-07-30
Adds spacing to docs

matthewlyle:
This adds basic sizing tokens and information on our spacing scale and why you'd use it to the docs.


docs/catalog/guidlines/spacing.md
```diff
@@ -0,0 +1,84 @@
...
+  <div style="display: inline-block; margin-right: 32px;">
+  	<div style="width: 64px; height: 64px; background: #000"></div>
+  	<p style="font-size: 12px; padding: 8px 0">x8</p>
+  </div>
+</div>
+```
+
+```image
+plain: true
+src: "/spacing-example.png"
```
theinterned, line 47: It would be helpful if this illustration included the token name as well as the pixel value

docs/catalog/guidlines/spacing.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+## How to choose spacing
+There are two important factors to consider when spacing: **size** and **relatedness**.
+
+* Use less spacing inside smaller elements or between functionally related elements.
+* Use more spacing inside larger elements or between less functionally related elements.
+
+Note: .5x should mostly be used for spacing related items within an element, e.g a button's text and it's icon.
```
theinterned: is `.5x` the same as `half`?

docs/catalog/guidlines/spacing.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Tokens
+
+```table
+span: 3
+rows:
+  - Token: space.half
+    px: 4
+    rem: 0.25
```
theinterned, line 65: Nice! I like that you give `px` and `rem` values

theinterned (changes requested):
This looks very good. The formatting in general looks nice. I just had a few comments to clarify some content and an issue with how we name scales.

tokens/index.js
```diff
@@ -24,8 +24,19 @@ export const font = {
   }
 };
 
+export const space = {
+  half: '4px',
```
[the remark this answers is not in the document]
matthewlyle: I'd prefer that too. The dot notation is why I ended up with half, or it would have just been .5.

The issue with shifting the whole scale is that 8 really is the base unit. 8, 16, 24, 32, 48, 64. x1, x2, x3, x4, x6, x8. The half is an exception for extra tight spaces because the next number down would be 0. Pretending like 4 is our base unit would make the scale x1, x2, x4, x8, x10, x12, x16. One of the main reasons for a scale here is language, to have the designers thinking in 8s and being able to communicate this to developers. 

I'm open to another suggestion but I couldn't think of any that were more clear than half.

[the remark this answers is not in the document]
theinterned: After some discussion we have agreed to stick with `half` for now. We can always revisit the name later.

theinterned (approved):

PR #22, opened 2018-08-22
Fix readme links

matthewlyle:
We have to use absolute URLs on the homepage because it's also used as our Github readme. This change will make the links work in both places.


theinterned (approved):
so good

PR #24, opened 2018-08-23
Update package scripts

theinterned:
# Problem

1. There wasn't an easy and obvious way to jump into component development.
1. Documentation was laking on how to use the scripts.

# What I did

## Added to and updated package scripts
1. Repurposes the `yarn start` script to start up the components Storybook.
1. Adds a `yarn document` script to start catalog in dev mode.
1. Adds a `yarn all` script that starts all storybooks and build all packages in watch mode.
1. Renamed the "Getting Started" section of the docs to "Guides": This section actaully contains more then just onboarding material ... so a name change better represents its purpose

## Updated docs
1. Adds a "Guides / Quick start" that just tells you the bare minimum to clone the repo, install deps and run a dev server.
1. Adds a "Guides / Package Scripts" doc outlining the use of the scripts in the root package.

# To test

Run the scripts.

Review the Doc changes.

- [Guides / Quick start](https://deploy-preview-24--nulogy-design-system.netlify.com/guides/quickstart)
- [Guides / Set up](https://deploy-preview-24--nulogy-design-system.netlify.com/guides/setup)
- [Guides / Package Scripts](https://deploy-preview-24--nulogy-design-system.netlify.com/guides/scripts)


matthewlyle:
scripts work and the docs look good. do you think we should have something about yarn all in the quick start guide? or did you want to make sure people who did yarn all read the entire set up? approved regardless as this is not a blocker.

matthewlyle (approved):

theinterned:
As far as I know running `yarn all` is not typical behaviour. I would prefer us having a couple of independent scripts you can fire off for different dev scenarios. This is just because I'm worried about flooding the cpu with a bunch of processes, but we should balance that against the pain we feel juggling different scripts. 

Are you using `yarn all` @matthewlyle  ? If not I am inclined to leave it out of the quick start and just point people toward the main development "personas".

Anyways, I will merge this now .. we can always change the scripts and docs if we find an issue

PR #25, opened 2018-08-23
enables jest tests in components package

theinterned:
# Problem

The Jest test config was broken in the @nulogy/components package.

# Solution

This PR restores some missing jest config files that were deleted from the CRA ejected config.

# To test

from the root run

```sh
yarn components test
```
There is a very basic test that should pass.


nikola-nulogy (approved):
Looks good.

PR #26, opened 2018-08-24
Css classes

matthewlyle:
I created corresponding css utility classes for each type token using the injectGlobal method and documented them both in Catalog and Storybook (in a new CSS story).


components/src/CSS/css.story.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
```
theinterned: Do you think it would make sense to have a storybook in the css package? I like this here, but it also would make sense to have the "tests" in the same package as the code.

What do you think @matthewlyle ?

css/src/typography.js
```diff
@@ -0,0 +1,27 @@
...
+  'font-size--subsection-title': css`font-size: ${font.size.sectionTitle}px`,
+
+  // Family
+  'font-family--regular': css`font-family: ${font.family.regular}`,
+  'font-family--mono': css`font-family: ${font.family.mono}`,
+
+  // Weight
+  'font-weight--bold': css`font-weight: ${font.weight.bold}`,
+  'line-height--regular': css`line-height: ${font.lineHeight.regular}`,
+}
```
theinterned, line 23: Would you be interested in generating this dynamically based on the keys in the tokens?

Something like:

```js
const buildMixins = name => Object.entries(font[name]).reduce((result, [key, value]) => ({ 
  ...result, 
  [`font-${name}--${key}`]: css`font-${name}: ${value}px` 
}), {});

const fontSize = buildMixins('size');
const fontFamily = buildMixins('family');
const fontWeight = buildMixins('weight');

Object.entries({ ...fontSize, ...fontFamily, ...fontWeight })
  .forEach(([className, rule]) => injectGlobal`.${className} { ${rule} }`);
```

I'm curious about your thoughts @matthewlyle @nikola-nulogy about the tursness of something like this.

One advantage of an approach like this is that if the keys were to change in the tokens, the css would update automatically.

On the other hand its much harder to read.

theinterned (commented):
everything looks good. See my question about code style

components/src/CSS/css.story.js
theinterned: After discussion with Matt we agreed to add storybook to the css workspace. I'm doing that.

theinterned: 👍 I added a typography story in the `css` and removed this one.

nikola-nulogy (approved):

css/src/typography.js
nikola-nulogy: I would prefer if we kept the original version.  It is more readable and familiar to a designer. This would make it less fragile in case designer is making the edits.

PR #29, opened 2018-08-31
Nds 228 design and build table cell and row

nikola-nulogy:
### What I did
- Styled Table component
- Created CreateRow, ActionCell, CreateRowCell, Button and TextInput (temporarily needed)
- Created Read Only, Edit Mode, and With Create Row sub-stories inside Table story

### To Test
Run storybook and access Table story


theinterned:
@nikola-nulogy you have a WIP commit in there

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
 import React from 'react';
 import styled from 'styled-components';
+import { colour, space, font, radius, shadow } from '@nulogy/tokens';
```
theinterned: `shadow` is unused. Please remove.

In fact no tokens should be used in this file ... those values should come from the theme.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+import QuietButton from '../Button/QuietButton';
 
-const borderStyle = ({ theme }) => `1px solid ${theme.colour.neutral[400]}`;
+const borderWidth = 1;
+
+const withoutBorder = value => value - borderWidth;
+
+const lineHeight = name => font.lineHeight[name] * font.size.medium;
```
theinterned: This is now used in two places ... it may be a good candidate for extraction ... perhaps to the Text component?

components/src/Tables/Table.js
theinterned: Please replace with an import of the tokens package (`import tokens from '@nulogy/tokens`) + a default `theme` prop.

See [`Link.js`](https://github.com/nulogy/design-system/blob/master/components/src/Link/Link.js#L23-L25) for an example the default `theme` prop.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
 
 export const Cell = styled.td`
-  padding: 1rem;
-  text-align: left;
-  border-bottom: ${borderStyle};
+  font-size: ${font.size.small}px;
+  line-height: ${lineHeight('smaller') / font.size.small};
```
theinterned: Do we actually want to set the font-size to be small and set a tighter line-height on tables? 

Quite a bit of the content in PackManager is tabular. This will have the effect of negating a lot of the roominess we were hoping to bring to the app.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  padding: ${space.x2}px ${space.x1}px;
+  &:first-child{
+    padding-left: ${space.x2}px;
+  }
+  &:last-child{
+    padding-right: ${space.x2}px;
+    width:72px; // not the right place
```
theinterned: This needs to go away.

We could set the width in JS based on the number of actions...

theinterned: is this width really specific to the last Cell? Or is this actually an `ActionCell` thing.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  }
 `;
 
-export const Row = styled.tr``;
+export const ActionCell = styled(Cell)`
+  padding: ${space.x1}px ${space.x2}px ${space.x1}px ${space.x1}px;
```
theinterned: a few comments here:

1. the right padding of `space.x2` is already set on the last-child td above and shouldn't be necessary.
2. The only things being overridden here are the top and bottom padding. I would prefer to see only those values set.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  display: flex;
+  justify-content: flex-end;
```
theinterned: why the flex here? would a simple `text-align: right;` not work?

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
-export const Header = styled.thead``;
+export const Row = styled.tr`
+  &:nth-child(odd){
+    background-color: ${colour.blue['200']};
+  }
+`;
 
-export const Body = styled.tbody``;
+export const Header = styled.thead`
+  text-align: left;
```
theinterned: it might make sense for Cell and Header Cell to support a textAlign prop:

```js
<Cell textAlign="left | centre | right | numeric">Foo</Cell>
```

with a `defaultProp.textAlgin` of `"left"`;

theinterned: ... the "numeric" property is stretching it, but could used to right align + switch to tabular numbers.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+`;
 
 export const HeaderCell = styled(Cell.withComponent('th'))`
   font-weight: bold;
-  border-bottom-width: 2px;
+  background-color: ${colour.white}; // sloppy
```
theinterned: is this needed? why the comment?

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+`;
+
+export const CreateRow = styled.tr`
+  background-color: ${colour.blue[800]};
+`;
+
+export const CreateRowCell = styled(HeaderCell.withComponent('th'))`
```
theinterned: 1. I am surprised that this is a `HeaderCell`, why not a `td`?
2. `HeaderCell` is already a `th`, no need to call `withComponent` again.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  font-weight: ${font.weight.normal};
+  padding: ${space.x1}px ${space.x1}px;
```
theinterned: Is this really a `CreateRowCell`? Maybe more generically it's an `InputCell`

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+`;
+
+export const Button = styled(QuietButton)`
```
theinterned: It would help me understand this Button if it was defined closer to the `ActionCell`

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  padding: ${withoutBorder(space.half)}px;
 `;
 
+export const TextInput = styled.input`
+  height:24px;
```
theinterned: 🤔

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+    background: ${colour.white};
+  }
+`;
+
+export const Body = styled.tbody``;
+
 export const Table = styled.table`
   border-collapse: collapse;
-  border-top: ${borderStyle};
+  table-layout: fixed;
```
theinterned: do we want to go fixed layout by default?

theinterned: it would make the sizing of the `ActionCell` a lot simpler for example if we didn't go to fixed layout by default.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
...
+  width: 100%;
 `;
 
 const DefaultTable =  styled(Table)``;
 
 Object.entries({
```
theinterned: I'm not sure about this API of exposing all the components as keys of the Table component.

theinterned (changes requested):
Overall the table is looking great! I left quite a few comments.

components/src/Tables/Table.js
```diff
@@ -1,38 +1,98 @@
 import React from 'react';
 import styled from 'styled-components';
+import { colour, space, font, radius, shadow } from '@nulogy/tokens';
+import QuietButton from '../Button/QuietButton';
 
-const borderStyle = ({ theme }) => `1px solid ${theme.colour.neutral[400]}`;
+const borderWidth = 1;
+
+const withoutBorder = value => value - borderWidth;
```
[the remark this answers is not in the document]
theinterned: also it is confusing to use the word `borderWidth` in this context as it is meant to apply to the `Table.Button` but I would expect it to indicate a border width for the Table or Cells.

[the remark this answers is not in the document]
theinterned: In cases where you are declaring consts for a side concern, it would be good to call that out: `buttonBorderWidth` for example.

theinterned:
I would also like to see some documentation in Catalog please.

components/src/Tables/Table.js
nikola-nulogy: I picked small because regular looked way to different from the current state. Currently table font size is 12px. I agree we should address this in the next release but for the 21st I would keep it as is.

components/src/Tables/Table.js
nikola-nulogy: This value is specific to the number of items in ActionCell. I tried solving this through ActionCell but it did not work out for me. Let's pair on this one.

components/src/Tables/Table.js
nikola-nulogy: Just a temporary solution until we make Input component. Entire TextInput is just a placeholder at the moment.

nikola-nulogy:
I will be doing documentation as a separate story.

nikola-nulogy:
I have added documentation and resolved all the issues I had. Time to have another look.

theinterned (approved):
Looks good! Tables are here! 

I'm going to cut another release ... I'd say this is breaking so I'm going to bump to `v0.1.5`

PR #42, opened 2018-09-13
NDS 280 build basic datalist

theinterned:
# Problem

We need a component to show key-value data in a multi-column list. This is used a lot in our products.

# Solution

I built a data-list component that supports the following features:

1. Pass data via a 2D array: `[[ key, value ], [ key, value ], ... ]` and it renders it!
2. Responsive columns. By default this is: 1 column on small screens, 2 on medium screens and 3 on larger screens.
3. Number of columns can be customized via a `columns` prop:
   - non-responsive pass an `int`: `columns={5}`
   - responsive pass an `Array`: `columns={[5, 7, 10]}`
4. Supports custom rendering via [child-as render-prop](https://reactjs.org/docs/render-props.html) that receives `key` and `value` args.

## The minimal api looks like: 

```jsx
<DataList data={[
  [ 'key 1', 'value 1'],
  [ 'key 2', 'value 2'],
]} />
```

## Customizing the columns:

```jsx
// 10 columns on small screens, 20 columns on medium screens, 30 columns on larger screens
<DataList data={data} columns={[10 , 20, 30]} />

// non-responsive
<DataList data={data} columns={5} />
```

## Custom render:

```jsx
<DataList data={data}>
  {(key, value) => (
    <React.Fragment>
      <h1><b>"{key}"</b> is in an <code>h1</code> tag</h1>
      <p><b>"{value}"</b> is in a <code>p</code> tag</p>
    </React.Fragment>
  )}
</DataList>
```

## the Column filling algorithm

I have coded it so that columns fill "greedily", meaning that columns to the left will fill up before the remainder are added to the last column.

This is done using flex box and `grid-template-rows`. I take the `Math.ceil( items / columns )` and render that many rows. Then I flow the items into the rows using `grid-auto-flow: column dense`. This causes the items to fill the columns as densely as possible.

### BUG

There is a display bug though where certain combos of columns / item #s don't render all the columns. [I have created a story to illustrate this.](https://github.com/nulogy/design-system/pull/42/files#diff-711721ceffaf9911d73f7193d839bde8R43)

# What else did I do?

## 1. Added breakpoints and media queries to [`tokens.js`](https://github.com/nulogy/design-system/pull/42/files#diff-88c843709e43880fb7441947ae33338eR113)

There are 2 breakpoints and I lifted them from [Zurb](https://foundation.zurb.com/sites/docs/media-queries.html)

```js
{
  medium: 640,
  large: 1024,
}
```

[...]

PR #43, opened 2018-09-14
Nds 282 base font size correction

nikola-nulogy:
What I Did

Changed base font size on Button and Table components

### DIFF
--- a/tokens/src/tokens.js
+++ b/tokens/src/tokens.js
@@ -1,5 +1,3 @@
-import styled, { css } from 'styled-components';
-
 export const borderRadius ='3px';
 
 export const font = {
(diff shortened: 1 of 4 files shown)


theinterned (changes requested):
There is a question about line heights we need to address ...

matthewlyle (approved):

tokens/src/tokens.js
```diff
@@ -22,11 +24,24 @@ export const font = {
...
+    },
+    ratio: {
+      smaller: 1,
+      small: 1.2,
+      medium: 1.5
+    }
   }
 };
 
+export const fontMetrics = (size = 'medium', target = 'medium') => ({ theme = { font } }) => css`
```
theinterned: I'm not sure if `@nulogy/tokens` should be exporting helper functions ... We should aim to keep it as simple as possible and as close to pure json data (even though its actually not technically conformant json atm).

I'm also pretty sure that `@nulogy/tokens` doesn't actually have `styled-components` in its package json. Nor should it.

This should go someplace else: Either a utils folder of the `@nulogy/components` package, or better yet, a new shared package `@nulogy/theme` perhaps?

There is a mediaQuery helper function introduced in PR #42 that would go nicely in that package as well.

theinterned (changes requested):

theinterned (approved):

PR #47, opened 2018-09-17
Nds 283 integrating icon component in button

nikola-nulogy:
### What I Did

Added Icon component to Button story
Styled Button component with Icon component
Reduced horizontal button padding (looks better with icon only buttons).


components/src/Button/Button.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import styled, { css } from 'styled-components';
 import { colour, space, font, radius, shadow } from '@nulogy/tokens';
+import { Icons } from '../Icons';
+const {Icon} = Icons;
```
theinterned: These imports are not used. remove them.

theinterned (changes requested):

components/src/Button/Button.js
theinterned: actually I'm in this PR now, so I'll do it

components/src/Button/Button.js
```diff
@@ -122,6 +136,9 @@ const Button = styled.button`
     pointer-events: none;
   }
 
+  * {
```
theinterned: I'm going to update this PR to wrap the children in a span to make this stying a little nicer

theinterned (changes requested):

components/src/Button/Button.js
```diff
@@ -92,7 +92,13 @@ const size = ({ size = 'medium' }) => (({
   `
 })[size]);
 
-const Button = styled.button`
+const Wrapper = styled.span`
```
theinterned: @nikola-nulogy I added this wrapper component to children when there are more than one (ie I don't add it to a single icon on its own or a button with only text).

The alignment and size of the button with an icon seems wrong though.

If you run storybook locally, then check out these stories:

[📖 With an icon and text](http://localhost:8080/?selectedKind=Buttons%20%2F%20Button&selectedStory=With%20an%20icon%20and%20text&full=0&addons=0&stories=1&panelRight=1&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel)

the alignment seems okay, but my inspector is showing this button as `33px` (I didn't check the dimensions for the other button sizes).

screen shot 2018-09-18 at 1 09 30 pm

[📖 With an icon](http://localhost:8080/?selectedKind=Buttons%20%2F%20Button&selectedStory=With%20an%20icon&full=0&addons=0&stories=1&panelRight=1&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel)

The size is off again: I read `30px`, and the icon is not aligned.

screen shot 2018-09-18 at 1 10 33 pm

theinterned (changes requested):
@nikola-nulogy I wrapped the individual children in a Wrapper span, and used that to add some space between each Wrapper, but there seems to be something off about the heights of the buttons. It looks like perhaps line-height is missing / not properly set. I wonder if you can have a look at that.

Thanks,

theinterned (approved):

PR #50, opened 2018-09-18
WIP: Nds 286 styling dl

nikola-nulogy:
What I Did

Style definition list in a controversial manner. If it passes table header can be styled the same way.

### DIFF
--- a/components/src/DataList/DataList.js
+++ b/components/src/DataList/DataList.js
@@ -28,7 +28,7 @@ export const List = styled.dl`
   ${ columns }
 `;
 List.displayName = 'NDS.DataList.List';
-List.defaultProps = { 
+List.defaultProps = {
   ...defaultProps,
   count: 1,
   columns: [1, 2, 3],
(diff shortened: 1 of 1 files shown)


components/src/DataList/DataList.js
```diff
@@ -44,7 +44,12 @@ List.propTypes = {
 
 export const Key = styled.dt`
   ${({theme}) => css`
-    font-weight: ${theme.font.weight.bold};
+    //font-weight: ${theme.font.weight.bold};
```
theinterned: remove

components/src/DataList/DataList.js
```diff
@@ -44,7 +44,12 @@ List.propTypes = {
...
+    font-size: ${theme.font.size.small}px;
+    color: ${theme.colour.neutral[600]};
+    text-transform: uppercase;
+    letter-spacing: .075em;
```
theinterned: I think this needs to be tightened up a bit

components/src/DataList/DataList.js
```diff
@@ -44,7 +44,12 @@ List.propTypes = {
...
+    line-height: 1.413; // related to https://github.com/nulogy/design-system/pull/43#discussion_r218503006
```
theinterned: we should definitely consider exporting the utils.

theinterned (changes requested):

theinterned (approved):

PR #57, opened 2018-09-25
Nds 297 create row button style broken

nikola-nulogy:
What I did

- Styled the button in CreateRow
- Replaces svgs in Table story with icon component
- Added hover animation to Table Button (Let's discuss)
- Added a Ripple effect to primary buttons (Let's discuss)

### DIFF
--- a/components/src/Button/Button.js
+++ b/components/src/Button/Button.js
@@ -12,9 +12,9 @@ const smallButtonPaddingY = 1;
 
 const themeProp = { theme: { font } };
 
-const iconSizeRatio = 
-  (fontSize = 'medium', lineHeight = 'medium') => 
-  ({ theme }) => 
+const iconSizeRatio =
+  (fontSize = 'medium', lineHeight = 'medium') =>
+  ({ theme }) =>
     (lineHeightRatio(fontSize, lineHeight)(themeProp) * iconSizeRatioForLineHeight(lineHeight)(themeProp));
 
 export const buttonReset = css`
@@ -28,16 +28,32 @@ export const buttonReset = css`
   cursor: pointer;
 `;
 
-export const buttonColours = (mainColour = colour.blue.base, ) => css`
+export const buttonColours = (mainColour = colour.blue.base) => css`
   color: ${mainColour};
   background-color: ${colour.white};
   border-color: ${mainColour};
+`
+
+export const Ripple = (mainColour = colour.blue[700]) => css`
+  background-position: center;
+
+  &:hover {
+    background: ${mainColour} radial-gradient(circle, transparent 1%, ${mainColour} 1%) center/15000%;
+    transition: all 0.75s;
+  }
+
+  &:active {
+    background-size: 100%;
+    transition: background 0s;
+  }
 `
 
 export const primaryButtonColours = (mainColour = colour.blue.base) => css`
   color: ${colour.white};
   background-color: ${mainColour};
   border-color: ${mainColour};
+
+  ${ Ripple(mainColour) }
 `
 
 const type = ({ type = 'button' }) => (({
@@ -60,6 +76,7 @@ const type = ({ type = 'button' }) => (({
 
   submit: css`
     ${primaryButtonColours()}
+    border-color: transparent;
 
     &:hover {
       ${primaryButtonColours(colour.blue[700])}
@@ -159,9 +176,8 @@ const StyledButton = styled.button`
 
   &:active {
     border-style: solid;
-    box-shadow: none;
+    //box-shadow: none;
     outline: none;
-    transform: scale(0.98);
     transition: .05s  ease-in;
   }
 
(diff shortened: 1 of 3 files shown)


components/src/Button/Button.js
```diff
@@ -159,9 +176,8 @@ const StyledButton = styled.button`
 
   &:active {
     border-style: solid;
-    box-shadow: none;
+    //box-shadow: none;
```
theinterned: delete

components/src/Tables/Table.js
```diff
@@ -25,6 +26,54 @@ export const Cell = styled.td`
...
+    ${IconWrapper} > svg{
+      height: 100%;
+      width: 100%;
+    }
+
+    &:hover{
+      color: ${theme.colour.neutral[600]};
+      transform: scale(1);
+    }
+    &::before {
```
theinterned, line 62: this is a lot of wiz-bang to only appear in the context of a table. Why is the table button so different form other buttons?

components/src/Tables/Table.js
```diff
@@ -33,31 +82,13 @@ export const ActionCell = styled(Cell)`
     white-space: nowrap;
     vertical-align: middle;
     & > button:not(:last-child) {
```
theinterned: This shouldn't be a `button` selector should it? Perhaps `${Button}`?

components/src/Tables/Table.js
```diff
@@ -33,31 +82,13 @@ export const ActionCell = styled(Cell)`
...
-      margin-right: ${theme.space.half}px;
```
theinterned, line 36: Have you considered the usbalitly of adding a large non-interactive space between buttons? I often find it easier to use button-groups where the buttons themselves have a large target zone, while the space between buttons is small.

components/src/Tables/Table.js
```diff
@@ -25,6 +26,54 @@ export const Cell = styled.td`
...
+    padding: ${theme.space.half}px;
+    border-width: 0;
+    border-radius: 50%;
+    transition: none;
+    color: ${theme.colour.neutral[600]};
+    ${IconWrapper}{
+      width: 1.5em;
+      height: 1.5em;
+    }
+    ${IconWrapper} > svg{
```
theinterned, line 58: Can we export and target the`Svg` component directly?

theinterned: As it is, this selector is overly-tied to internal knowledge of the Icon components markup.

components/src/Tables/Table.js
```diff
@@ -84,6 +115,12 @@ export const CreateRow = styled.tr`
     ${ ActionCell } {
       padding-top: ${theme.space.x2}px;
       padding-bottom: ${theme.space.x2}px;
+      ${ Button } {
```
theinterned: 🚨 These nested selectors are against the law. They are two deploy nested and don't leak styles form other componets across each other. 

Please remove them and use the parent-nested selector pattern instead: See the section "sass parent selectors" and "explicit rules around sass nesting" in http://bradfrost.com/blog/post/css-architecture-for-design-systems/

components/src/Tables/Table.js
```diff
@@ -84,6 +115,12 @@ export const CreateRow = styled.tr`
     ${ ActionCell } {
```
theinterned: When is an `ActionCell` not a child of a Row? Why are these styles not defined on the `ActionCell` directly?

theinterned (changes requested):
I have not pulled and looked at these changes yet, but I have left some comments just on the code

components/src/Tables/Table.js
nikola-nulogy: It is the only circular (square) shaped button. Here I suggested a special hover treatment for this button and that treatment only works for icon only or circular (square) shaped buttons.

components/src/Tables/Table.js
nikola-nulogy: I have and I agree with that. In this case, buttons are squeezed a bit more to save some space and make "responsive demo" more attractive. Having said that I will add extra space now that Shah demo is over.

components/src/Tables/Table.js
nikola-nulogy: And if we decide that this is the way to go we can consolidate this style with the rest of the buttons. This was more an exploration that anything else.

components/src/Tables/Table.js
nikola-nulogy: FIxed.

components/src/Tables/Table.js
nikola-nulogy: I believe this is fixed as a result of fixing the previous comment

components/src/Button/Button.js
nikola-nulogy: I removed ripple effect and removed comment here since this line of code is needed in absence of ripple effect.

components/src/Tables/Table.js
nikola-nulogy: Can we extract this one from this pull request?

theinterned: I do not believe we can. I do not want to allow poor coding practices to pile up in our code base and accrue debt.

PR #59, opened 2018-09-26
Nds 309 document line height 

nikola-nulogy:
What I Did

- Created Vertical Rythm story under Type story.
There is an ugly background image that needs to be replaced with the image I 've created but need help with getting it into StoryBook
- Adjusted line-heights and bottom margins of Type components
- Made changes to Typography and Type documentation to include line height
- Added line height tokens to token documentation
- Documented fontMetrics utility function under typography.
This isn't the most appropriate place for this but until we solidify utils strategy it is probably the best place for it.
- Adjusted the line height of dd to fit on the vertical grid


theinterned:
@nikola-nulogy this has been open for several days ... it looks like you mention needing some help. What are the next steps?

nikola-nulogy:
> @nikola-nulogy this has been open for several days ... it looks like you mention needing some help. What are the next steps?

We are having a meeting to go over the work that was done here.

Actually yes, I needed help with getting static assets into storybook. This is not a blocker for today's meeting since I found the workaround but is something that we should find a solution for.

matthewlyle (approved):

PR #60, opened 2018-09-26
moves jest and babel-jest out of direct deps

theinterned:
Quick fix to move Jest deps out of direct dependencies and into dev-dependencies to reduce bundle size.


theinterned:
Closing this in favour of #61

PR #62, opened 2018-10-03
Nds 277 document icon component

matthewlyle:
This PR adds basic documentation for Icon in Visual Design and Component section of docs.


docs/catalog/visual_style/iconography.md
```diff
@@ -1,3 +1,46 @@
...
+    <Icon name="menu" />
+    <Icon name="save" />
+    <Icon name="search" />
+    <Icon name="site" />
+    <Icon name="unlock" />
+    <Icon name="user" />
+</div>
+```
+
+If your interface requires an icon not listed here, please post a message in the [#design-system](slack://channel?id=CBAFQ4X7X/) slack channel.
```
theinterned, line 28: Perhaps a PR flow would be appropriate here? 

Then whoever could fork the component, add the icon they need and submit a PR and not be blocked.

theinterned: "
If your interface requires an icon not listed here, please feel free to fork the [Icon component on Github](https://github.com/nulogy/design-system/tree/master/components/src/Icons) and open a pull request. 

The name should be a camel-case version of the icon name shown [here](https://material.io/tools/icons/). For example `add_to_queue` becomes `addToQueue`.
"

docs/catalog/visual_style/iconography.md
```diff
@@ -1,3 +1,46 @@
...
+
+## Sizes
+Icons can be 8, 16, 32, 48 or 64 pixels in size.
```
theinterned: These sizes aren't 100% accurate. For example, the icon is currently `20px` for a `16px` font-size.

The icon is sized as `1.25em`.

theinterned: also, is contextual sizing enough? Or would you like a `size` prop for the icon ... maybe it could multiply the `1.25` ratio by `props.size`?

nikola-nulogy:
"Try to avoid using icons without labels unless the meaning is ubiquitous, like print."

What do you think about saying "Avoid ..." rather than "Try to avoid ..."? I think it's stronger, I believe that what's being used so far.

nikola-nulogy:
"They will be the same size and colour as their surrounding text."

Icon can be surrounded not by text only in which case it would inherit the colour and the size from that element (Button for example) so it may be more precise to say "... as parent element".

Right now icon inherits the size of the surrounding font and makes it 25% larger. It is true that the size is being inherited and adjusted but the sizes don't match. We can say something like "They will be the same colour, and their size will be based of their parent element. The icons are 25% larger than parent element text-size." That is what is happening right now. we can discuss if this is working for us or not at some point.

nikola-nulogy (approved):

PR #70, opened 2018-10-12
Update type styles in docs

matthewlyle:
* Updates type sizes and line heights to new sizes and line heights
* Moves Catalog CSS out of html page to an actual CSS file


nikola-nulogy:
I guess this doesn't include paragraph styles. I am ok will dealing with this in another PR.

nikola-nulogy (approved):

PR #77, opened 2018-10-22
Table UX guidelines

matthewlyle:
This PR adds a basic UX Guidelines section to the table docs that explains best practices regarding cell alignment and truncation.

### DIFF
--- a/docs/catalog/components/tables.md
+++ b/docs/catalog/components/tables.md
@@ -121,4 +121,5 @@ Columns should be aligned based on their content types, according to the followi
 * Do not center anything
 
 ### Truncation
-When possible, wrap content instead of truncating. This helps avoid ambiguity, especially with content that starts with the same letters. If necessary, consider adding a Tooltip.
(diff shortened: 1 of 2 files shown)


neilio (commented):
Should we have a separate rule for columns that have checkboxes? I think they might work best centred.

nikola-nulogy:
Should we consider making truncation a UX decision rather than saying "When possible ..."? It feels that that heavily depends on the context. Maybe we can focus here on providing guidelines if truncation is necessary. In some cases, it makes sense to truncate at the end but not always. Maybe this is something that we can talk about here. Can we provide some kind of recommendation in terms of character number or number of lines? I understand that this is not a table-only concern and maybe there is a more appropriate place to get more in-depth with truncation?

matthewlyle:
@neilio Hm do you have an example of centering being best for checkboxes? I _think_ left aligned checkboxes work just fine?

nikola-nulogy (approved):

PR #82, opened 2018-12-19
Table documentation

nikola-nulogy:
I added a table component to our documentation site. I have tried something different here and instead of talking about the parts and anatomy I talked in terms of functionality.


matthewlyle:
I removed any references to things that don't exist yet as we can just add references to each new component as we make it. I also removed the front-end accessibility information and put it in a Confluence doc so we can put it in when we know where we want to put the front-end stuff.

matthewlyle (approved):

PR #83, opened 2018-12-20
updates netlify build settings

matthewlyle:
We changed the build scripts which means we have to change the build scripts netlify tries to run.


neilio (approved):
Looks good to me ¯\_(ツ)_/¯

PR #84, opened 2018-12-20
Fix netlify

matthewlyle:
Had to remove the reference to the old Link component as that's no longer working correctly and would cause the Catalog build to fail. This is fine because we don't actually need that document right now.


neilio (approved):
Looks good to me.

PR #95, opened 2019-01-10
adds icon component and icon stories

coostenbrug:
Creates Icon component + storybook stories for NDS by wrapping icon component from:
https://github.com/jxnblk/rmdi

Icon component should:
 - wrap rmdi components with NDS defined Icon "name" prop
 - accept props as documented from rmdi
 - inherit styles from parent components
 - align with surrounding text
 - meet accessibility requirements


matthewlyle:
When running `yarn build` I receive the following error:

```ERROR in ./src/Icon/Icon.js
Module build failed (from ../node_modules/babel-loader/lib/index.js):
SyntaxError: /Users/mattd/src/design-system/components/src/Icon/Icon.js: Unexpected token (38:4)

  36 |
  37 | const Icon = ({name, IconSvg = Svgs[name],...props}) => (
> 38 |     <IconSvg aria-hidden={!props.title} style={{top: (props.size/8)||"0.125em",position: "relative"}} size={props.size||iconSizeRatio} {...props}/>
     |     ^
  39 | );
```

This causes an error in storybook as well, ` TypeError: Object(...) is not a function`

Also, when writing a component that accepts many attributes like this it can be cleaner to read when indented like: 
```
const Icon = ({name, IconSvg = Svgs[name],...props}) => (
    <IconSvg 
        aria-hidden={!props.title} 
        style={{top: (props.size/8)||"0.125em",position: "relative"}} 
        size={props.size||iconSizeRatio} 
		{...props}
     />
);
```

coostenbrug:
Made Icon.js more readable as suggested and fixed this build issue you were getting. Was an issue with babel and fixed by adding the react preset to babel-loader.

coostenbrug:
Closing pull request as the work for Icon Component has moved to branch

PR #99, opened 2019-01-11
Nds 476 load stories dynamically

coostenbrug:
This makes changes to the css and component storybook config files so that all stories are loaded as long as they are in the correct path and names <name>.story.js 

This seems like a good solution for now so long as we are okay with all stories loading in alphabetical order, as that is the way the they are organized in the folder structure and therefore the order they are loaded. If we want to have more granular control we can either revert these changed and continue to add stories manually or we can look into a way to index the stories to appear in a desired order.


matthewlyle (approved):
Alphabetical is good, yeah.

PR #100, opened 2019-01-11
Typography margins and line heights

nikola-nulogy:
**What I did**
 - Added default margin-bottom values to existing typographic components
- Created Typography/Article story
- Added missing line heights for existing  typographic components (these values are hard-coded for now and we need to discuss how we want to handle them)
- Adjusted font weights for existing typographic components
- Gave typographic components appropriate tag instead using generic div/h1

**How to test**
- Run Typography/Article story in storybook

**Notes**
- margin-bottom from inline text component needs to be removed and I've created a [ticket](https://nulogy-go.atlassian.net/browse/NDS-487) for it.
- Margin resetting is also another thing that we should look at


matthewlyle:
For the line heights, what about adding a file called   `/size/lineHeight/lineHeight.json` with the following tokens:
```
{
    "size": {
      "lineHeight": {             
        "base"  : { "value": "24" },
        "pageTitle"  : { "value": "" },
        "sectionTitle"  : { "value": "" },
        "subsectionTitle"  : { "value": "" }
      }
    }
  }
```

We'd need to use pixel values instead of the unitless, since that folder gets processed as PX for now. I don't think this should be an issue, though. 

We can then add an object to theme.js called lineHeights that references them, and they can be referred to like anything else, e.g `theme.lineHeights['pageTitle']`.

matthewlyle:
And I'm not quite sure what you mean by "margin resetting", but fyi I have a ticket in JIRA (NDS-488) to handle adding 0 to our spacing scale so we can say m=0 and it not be 4px. Maybe that will solve the issue.

nikola-nulogy:
> theme.lineHeights['pageTitle']

I have created a new [ticket](https://nulogy-go.atlassian.net/browse/NDS-489) to allow handling of unitless values.

matthewlyle (approved):

matthewlyle:
For some reason the spacing tokens were duplicated in the tokens `/dist/` files. I fixed that and everything seems to be running fine.

PR #101, opened 2019-01-14
Nds 484 Icon Component

coostenbrug:
This is the completed Icon Component, exports the following:
Icon <-- Icon component that is expected to be used
Svg <-- Component containing the Svg that can be used when needed

Icon accepts space from styled-system, color/fill, title, and name to specify what svg to use

The svgs are stored in in components/src/Icon/icons.svg.

For this iteration of the Icon Component, custom sizes are not built-in and instead size of the icon always scales to the font-size of the parent. Custom sizes are possible but not build into the components API. It may be best to put a hold on this issue until we run into use-cases or get a better understanding of all of the ways this Icon component may be used.


coostenbrug:
Made the following changes:

 - svgs are stored separately rather than in a single file, paths are collected into icons.json. I have made a new ticket to build a script that converts all svgs into the json object.
 - Svg component is now the standard Icon component (it does not come with alignment)
 - Icon component is now called InlineIcon
     - This can be either for external use, internal use only (for use in other components), or removed entirely if we deem that it is not needed
 - Stories have been cleaned up significantly

PR #104, opened 2019-01-16
Updated README with directory change

neilio:
The instructions were missing the step to change directories before running `yarn` so I added it.


matthewlyle:
I moved the directory change into the clone line, since you wouldn't really want to ever not immediately go into that directory.

matthewlyle:
This has been taken into account in the new docs but this readme doesn't include this information anymore so isn't necessary.

PR #117, opened 2019-01-22
NDS-505 Icon Component Documentation

coostenbrug:
Adds docs page for icon modelled from the old iconography docs page. If you see any changes you think should be made to the wording or the spacing of elements on the page let me know. 

The icons displayed are dynamically loaded so they should display all icons from whatever version of the NDS components package will be loaded on the docs site without the need to manually add new icons


nikola-nulogy:
Content Guidelines are more Accessibility Guidelines. Also, 2nd line doesn't seem to be a guideline. I believe that it belongs to Related or Learn more section.

nikola-nulogy (approved):

PR #120, opened 2019-01-22
fixes bug where icon shrinks when window causes button to shrink

coostenbrug:
I noticed that the previous pr had a bug so I opened a new one here that resolves the issue properly


nikola-nulogy:
The size of the label in iconic button has turned to 1px.

PR #129, opened 2019-01-24
Nds 546 code list component

nikola-nulogy:
### What I did

- Created List and List Item components component
- Added List and updated Typography/Article story to storybook

### How to test

- Run LIst and Article stories in storybook


components/src/List/List.js
```diff
@@ -0,0 +1,36 @@
...
+import styled from "styled-components";
+import {
+  space, color, fontSize, fontWeight, lineHeight,
+} from "styled-system";
+import theme from "../theme";
+import ListItem from "./ListItem";
+
+const List = styled.ul`
+  list-style: none;
+  -webkit-font-smoothing: antialiased;
```
clemensp: Curious about `-webkit` style attributes in general: Should we be adding the attributes that target the other browsers as well?

components/src/List/List.js
```diff
@@ -0,0 +1,36 @@
...
+  margin: 0;
+  padding-left: ${theme.space[4]};
+  ${space}
+  ${color}
+  ${fontSize}
+  ${fontWeight}
+  ${lineHeight}
+
+  ${ListItem} {
+    margin-bottom: ${props => (props.compact ? "0" : props.theme.space[2])};
```
clemensp: This can be shortened a bit with: `${({compact, theme}) => `

components/src/List/List.js
```diff
@@ -0,0 +1,36 @@
...
+
+List.propTypes = {
+  ...space.propTypes,
+  ...color.propTypes,
+  spacing: PropTypes.bool,
+};
+
+List.defaultProps = {
+  mb: 4,
+  color: "inheretColor",
```
clemensp: inheret -> inherit?

Also, is this different from `currentColor`?

components/src/List/List.story.js
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import List from "./List";
+import ListItem from "./ListItem";
+
+storiesOf("List", module)
+  .add("List", () => (
+    <React.Fragment>
```
clemensp: Don't need `React.Fragment` if there is only one top-level node.

clemensp: Same for the stories below.

components/src/Validation/InlineValidation.story.js
```diff
@@ -1,23 +1,26 @@
...
 
 storiesOf("Inline Validation", module)
   .add("Inline Validation", () => (
     <InlineValidation message="Something has gone wrong">
-      <ul style={ { margin: 0, WebkitFontSmoothing: "antialiased" } }>
-        <li>Something has gone wrong.</li>
-        <li>Entry must be atleast 3 characters long</li>
-        <li><a href="https://nulogy.design/">See here</a></li>
-      </ul>
+      <List compact>
```
clemensp, line 11: Nice. 😃

clemensp (changes requested):
Some questions to address, but the component looks nice. 👍

components/src/List/List.js
matthewlyle: The firefox fontsmoothing is `-moz-osx-font-smoothing: grayscale;`

I wonder if we should be applying this globally though... it's not really a decision unique to List, rather we want all of our fonts to be antialiased. 

We could add it to `ThemeProvider.js` and have it come along when we set the font to Plex.

nikola-nulogy: I agree we should have this applied globally.

clemensp: Yeah. There's probably a library that provides these.
They usually come in sets of 3, right?

coostenbrug: Put this on our ThemeProvider. Ideally we can use a library to handle these attributes on different browsers so I made a new ticket to handle that.

components/src/List/List.js
coostenbrug: theme was imported already so I just removed props in props.theme.space[2]

components/src/List/List.js
coostenbrug: changed to currentColor

components/src/List/List.story.js
coostenbrug: fixed

clemensp (approved):
LGTM. 👍

PR #130, opened 2019-01-24
eslint round 2

clemensp:


components/src/Icon/Icon.story.js
```diff
@@ -86,7 +86,7 @@ This has a title attribute so it will be read by assistive devices.
       <p>
         <Icon name="user" />
         {" "}
-This doesn't have a title attribute, so it has aria-hidden set true instead.
+This doesn&apos;t have a title attribute, so it has aria-hidden set true instead.
```
matthewlyle: Is it possible to escape with a `\` instead of the entity code with eslint? It might be a little more readable.

matthewlyle (commented):
When I run `yarn icons` on this branch I'm getting: ```/Users/mattd/src/design-system/components/scripts/collect-icon-svgs.js:1
(function (exports, require, module, __filename, __dirname) { import fs from "fs";
                                                              ^^^^^^```

matthewlyle (approved):

PR #134, opened 2019-01-29
Add jest and storyshots

clemensp:


components/src/Box/Box.story.js
```diff
@@ -1,16 +1,14 @@
 import React from "react";
 import { storiesOf } from "@storybook/react";
-import { withInfo } from "@storybook/addon-info";
+// import { withInfo } from "@storybook/addon-info";
```
matthewlyle: this could be removed instead of commented out.

matthewlyle:
What is the point of the SpecHelper.js? Do we plan on putting things in it?

matthewlyle (approved):

components/src/Box/Box.story.js
clemensp: Removed.

clemensp:
> What is the point of the SpecHelper.js? Do we plan on putting things in it?

It's the place to put setup/configuration that is shared across specs.
There used to be a line in there, but I removed it because it wasn't needed.

I think the file is still nice to have to provide that central spot for setups, so I'll leave it.

PR #142, opened 2019-01-30
Nds 494 code toggle

coostenbrug:
Adds Toggle and Toggle Stories
This toggle supports animation, accessibility, and correctly works with focus

Use the story marked with <TEMP> to see the functionality of the toggle and its various props. (We can remove this story later if we don't want demonstrations of functionality like this in storybook)

the starting basis of the component came from here: 
https://github.com/aaronshaf/react-toggle/blob/master/src/component/index.js

### DIFF
--- a/components/src/Toggle/Toggle.js
+++ b/components/src/Toggle/Toggle.js
@@ -11,6 +11,20 @@ cursor: pointer;
 background-color: transparent;
 `;
 
+const focusedStyle = focus => {
+  if(focus) {
+    return(
+      {
+        outline: "none",
+        color: `${theme.colors.black}`,
+        borderColor: `${theme.colors.blue}`,
+        boxShadow: `0 0 3px ${theme.colors.blue}`,
+      }
+    );  
+  }
+  return;
+};
+
 const ToggleBG = styled.div`
 width: 64px;
 height: 40px;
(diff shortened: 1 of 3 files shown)


components/src/Toggle/Toggle.js
```diff
@@ -0,0 +1,160 @@
...
+const focusedStyle = focus => {
+  if (focus) {
+    return (
+      {
+        outline: "none",
+        boxShadow: `0 0 10px ${theme.colors.blue}`,
+      }
+    );
+  } else {
+    return false;
```
clemensp: What does false do?

clemensp (changes requested):
@coostenbrug : There's quite a bit going on in this one, so I'd like to pair with you for the review.

I checked it out in storybook, and the behaviour is nice. Looks great too!
Probably need @nikola-nulogy to review the box shadow, etc. =) 

On first pass, I see something that will have to be changed.
The standard API for react form elements is:
- value
- onChange
- defaultValue

You can also choose whether the form component is controlled or not:
https://reactjs.org/docs/forms.html#controlled-components

Let's work on those together.

I also wonder if we should simplify the Toggle to be just the pill, which can be an atom.
We could then introduce a ToggleField as a molecule, and compose it with `Field`, `Toggle`, and something that represents the value, possible just `Text`?

Looking forward to the discussion! :)

nikola-nulogy:
The disabled state should not have pointer cursor.

coostenbrug:
Closing and moving to the alt branch

PR #143, opened 2019-01-31
Form component

nikola-nulogy:
### Background

The purpose of Form component is to structure the form elements. There are 2 ways how to do that:
- With form sections* and,
- Without form sections

*Form Section is equivalent to fieldset in hmtl.

### What I did

- Created Form and Form Section component(s)
- Created Form story in stroybook
- Turned Field and Subheading component into style component so that it can be extended and styled through parent component

### Open questions

1. Naming: Form Section vs Fieldset vs Field Group component; Form vs Form Layout component.


components/src/Field/Field.js
```diff
@@ -25,6 +26,9 @@ const Field = ({
   </Label>
 );
 
+const Field = styled(BaseField)`
+`;
+
 Field.propTypes = {
```
clemensp: PropTypes have to be moved to `BaseField`

components/src/Field/Field.js
```diff
@@ -25,6 +26,9 @@ const Field = ({
   </Label>
 );
 
+const Field = styled(BaseField)`
+`;
```
clemensp: The template literal could go on the same line: ` const Field = styled(BaseField)`` `

components/src/Form/Form.story.js
```diff
@@ -0,0 +1,138 @@
...
+import { storiesOf } from "@storybook/react";
+import Field from "../Field/Field";
+import Input from "../Input/Input";
+import Form from "./Form";
+import FormSection from "./FormSection";
+
+storiesOf("Form", module)
+  .add("Form", () => (
+    <Form
+      title="New Profile"
```
clemensp: Can you try putting the attribute on the same line?
I'm trying to see which one's more readable.

clemensp (approved):
Looks good.
Included a couple of stylistic changes.

You will also have to update proptypes.

PR #150, opened 2019-02-05
Convert breakpoints to the object form

clemensp:


clemensp:
I'm pretty sure the only thing using breakpoints is the box story, to test responsiveness.

I checked docs as well, and there doesn't seem to be any.

coostenbrug (approved):

PR #168, opened 2019-02-08
[NDS-444] Decide on borders

matthewlyle:
Be prepared for a lot of words for a small PR.  Looking for either a +1 or discussion if any of my assumptions were incorrect. 

#### Things this PR does: 
1) Renames border radius to make more sense (`radius_border` instead of `radius_background`) 
2) Turns borderRadius theme.js array into an object 
3) Adds `radius_border_circle` for 50% 
4) Confirms that 4px is our normal medium border radius token, as that's used in both our inputs and buttons 
5) Keeps "small" as 2px though it's not being used anymore and we should keep an eye on its usefulness (see next section for why I didn't want to make it 1px) 

#### Things this PR doesn't do:
1) Add a token for 1px -- when we use 1px it's for a very specific purpose (like inside the checkbox) and it needs to be 1px, so there's no use mapping it to a token name as it should never change and it's already very descriptive. 
2) Add a large --- it feels strange to only have small and medium but since we don't know where we need a large yet I didn't add one. 8px sounds like it would make sense and match other systems doing something like 1,2,4 and 8, but without a use case it seemed like a premature optimization
3) Add semantic tokens - I was considering creating something like `border_radius_input` and `border_radius_button` but after a lot of consideration I decided not because 
	a) while there are a lot of articles about semantic tokens, most design systems I looked at don't use them (Polaris, carbon, Priceline, 24ways, enviro...)
	b) we aren't planning on building css-based components so I don't see a huge amount of use in semantic tokens since they only get used in one place anyway 
	c) the border references only occur in a couple places, and all reference "medium" which seems to make sense already
	d) we can always create semantic tokens later when we realize it's actually an issue or have a strong reason to do so 
	All that being said, I'm still open to adding semantic tokens, I just want to make sure we don't layer in unhelpful complexity.

### DIFF
--- a/components/src/Input/Input.js
+++ b/components/src/Input/Input.js
@@ -31,10 +31,9 @@ const getBorderColor = props => getInputStyle(props).borderColor;
 
 const Input = styled.input`
     width: 100%
-    color: ${getColor}; 
+    color: ${getColor};
     border: 1px solid;
-    border-color: ${getBorderColor}; 
-    border-radius: ${theme.radii[1]};
+    border-color: ${getBorderColor};
     border-radius: ${theme.radii.medium};
     padding: ${subPx(theme.space[2])};
     font-size: ${theme.fontSizes[1]};
@@ -45,7 +44,7 @@ const Input = styled.input`
     &:focus {
         outline: none;
         color: ${theme.colors.black}
-        border-color: ${theme.colors.blue}; 
+        border-color: ${theme.colors.blue};
         box-shadow: 0 0 3px ${theme.colors.blue};
     }
 `;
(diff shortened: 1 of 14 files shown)


components/src/Input/Input.js
```diff
@@ -35,6 +35,7 @@ const Input = styled.input`
     border: 1px solid;
     border-color: ${getBorderColor}; 
     border-radius: ${theme.radii[1]};
```
coostenbrug: Forgot to remove old border-radius reference

coostenbrug (approved):
+1 to everything said about 1px token, large token, and semantic tokens.
Looks nice a clean having these 3 options for radii for now.

Found a one-line issue, but other than that all good to merge once fixed!

nikola-nulogy (approved):

PR #170, opened 2019-02-08
NDS-579 Border-box for buttons

coostenbrug:
Intent: review for merge

The original purpose of this ticket was to explore if I could remove the subPx function and use border-box for inputs and buttons. My findings are that since buttons need to wrap and change height, height cannot be set meaning that border-box is not useful for this application. 

Things that are done in this ticket:
 - cleaned up some styling on Button 
 - adds focus to all buttons (the styling is the same as the button's hover)

### DIFF
--- a/components/src/Button/Button.js
+++ b/components/src/Button/Button.js
@@ -76,9 +76,7 @@ const Button = styled(BaseButton)`
     display: inline-flex;
     justify-content: center;
     align-items: center;
-    -webkit-font-smoothing: antialiased;
     font-weight: ${props => props.theme.fontWeights[2]};
-    border: 0;
     text-decoration: none;
     vertical-align: middle;
     line-height: ${props => props.theme.lineHeights.base};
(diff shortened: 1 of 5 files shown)


components/src/Button/Button.js
```diff
@@ -95,9 +93,17 @@ const Button = styled(BaseButton)`
     &:hover {
       background-color: ${props => (props.disabled ? null : props.theme.colors.lightBlue)};
     }
-    &:focus {outline: none;}
-    &:active {transform: scale(0.98); transition: .2s ease-in;}
-    &:disabled {opacity: .5;}
+    &:focus {
```
matthewlyle: this could all be combined with just `&:hover, &:focus {}`

coostenbrug: @matthewlyle After looking at it more - what do you think of using the hover styling as the focus? Mainly because it kinda breaks the illusion that the component is a physical button. ie. when you hover -> click the button they styling appears the same as when you hovered almost making it feel like that button hasn't "released" and is still pressed into the screen. Very subjective of course but let me know your thoughts.

nikola-nulogy (approved):

PR #173, opened 2019-02-08
Nds 578 add disabled prop to text component

coostenbrug:
Intent: Review to prepare for merge (one change must still be made)

This PR add the disabled prop to the Text component and a matching story
The disabled text is used in Toggle, Checkbox, and Radio.

Currently disabled Text is set at 50% opacity. This allows the for a color to be selected to contrast the background first, so that the disabled text properly contrasts as well (see story).
This decision at 50% may be altered and is the last question before this PR can be merged.

### DIFF
--- a/components/src/Checkbox/Checkbox.js
+++ b/components/src/Checkbox/Checkbox.js
@@ -70,7 +70,7 @@ const BaseCheckbox = props => {
       <CheckboxWrapper disabled={ disabled }>
         <CheckboxInput type="checkbox" { ...props } />
         <VisualCheckbox disabled={ disabled } checked={ checked } />
-        <Text disabled={ disabled }mb={ 0 }> {labelText} </Text>
+        <Text disabled={ disabled } mb={ 0 }> {labelText} </Text>
       </CheckboxWrapper>
     </Box>
   );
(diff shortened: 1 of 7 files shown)


components/src/Checkbox/Checkbox.js
```diff
@@ -71,7 +70,7 @@ const BaseCheckbox = props => {
       <CheckboxWrapper disabled={ disabled }>
         <CheckboxInput type="checkbox" { ...props } />
         <VisualCheckbox disabled={ disabled } checked={ checked } />
-        {labelText}
+        <Text disabled={ disabled } mb={ 0 }> {labelText} </Text>
```
clemensp, line 73: Why do we need `mb={ 0 }`? I see it throughout this review.
I wonder if the default should be 0. To follow the "no margin/padding on atoms" guideline.

clemensp (commented):
I'm just leaving general feedback, since I'm not the best person to comment on how opacity is handled. =)
To me, it looks good, except the resetting of margins. :)

components/src/Checkbox/Checkbox.js
nikola-nulogy: Opacity looks good for now. Maybe it will be slightly adjusted when we implant this somewhere.

coostenbrug: Our current Text component uses a margin bottom for spacing. The intended fix is in this ticket: https://nulogy-go.atlassian.net/browse/NDS-539 Where we discussed that Text should not have mb, and that there should be a [Paragraph?] component that is essentially Text with a default mb. That way we can remove all these instances of mb={ 0 }. For the sake of this PR we can ignore it and that work should be handled in that ticket

PR #174, opened 2019-02-11
Add a select component

clemensp:


components/src/Form/Form.story.js
```diff
@@ -122,6 +129,11 @@ storiesOf("Form", module)
           >
             <Input />
           </Field>
+          <Field
+            labelText="Project status"
```
coostenbrug: can move this prop to 1 line

components/src/Select/Select.js
```diff
@@ -0,0 +1,171 @@
...
+
+const getBorderColor = (errored, isOpen, disabled, isFocused) => {
+  if (errored) { return theme.colors.red; }
+  if (disabled) { return theme.colors.lightGrey; }
+  if (isOpen || isFocused) { return theme.colors.blue; }
+
+  return theme.colors.grey;
+};
+
+const SelectBox = styled.div([], ({ error, isOpen, disabled }) => ({
```
coostenbrug: I assume eslint will catch this, but unused input args here

coostenbrug:
Bugs:
- z-index of the menu should probably be increased so that it is always on top of other components 
- currently behaviour of if the menu is open and there are two select's is inconsistent: if top is open and bottom is closed, clicking bottom will close top. If top is open and bottom is closed, and bottom is opened, top remains open. EDIT: further investigation reveals this may be something related to another bug with defaultIsOpen. It seems like defaultIsOpen does not close the menu when an option is selected (both through clicking or with tab+enter).

components/src/Select/Select.js
```diff
@@ -0,0 +1,171 @@
...
+          <SelectBox { ...getToggleButtonProps({ disabled, error, isOpen }) }>
+            <Input
+              { ...getInputProps({ disabled, error, isOpen }) } aria-required={ required }
+              value={ selectedItem && selectedItem.label } placeholder={ placeholder }
+              readOnly
+            />
+            <ToggleButton isOpen={ isOpen } />
+          </SelectBox>
+          {
+            isOpen
```
coostenbrug: consider using && for conditional rendering instead of turnery (looks nicer IMO, but others may disagree!)

clemensp: Fixed.

components/src/Form/Form.story.js
clemensp: Fixed.

components/src/Select/Select.js
```diff
@@ -0,0 +1,175 @@
...
+              && (
+                <Menu { ...getMenuProps({ error }) }>
+                  {
+                    isOpen
```
coostenbrug: You can probably drop this inner conditional render, the outer one is already based on isOpen anyway

PR #176, opened 2019-02-12
Nds 539 remove margins on text

coostenbrug:
Intent: review to merge

This PR:
 - Remove the default mb value from the Text component
 - Updates other components to no longer need to set mb={0} manually
 - Updates the docs with a temporary solution mimicking the old Text component with mb={4}
Further notes/questions:
 - I tried to reduce the opacity for Text components when set to disabled using the polished library and set the color instead of setting the opacity. This however was not successful as this method makes it not possible to still inherit color as a style and optionally use a default. Text component should inherit color as a style as it is often a flexible component used within other components that can have color changes.
 - We need a true method of setting recommended margin bottoms to type components including Text.


nikola-nulogy:
Headings have turned into paragraphs rather than being h1, h2, h3.

PR #191, opened 2019-02-19
Nds 603 input clickable area

coostenbrug:
Intent: Review to merge

This PR extends the hitbox on the Toggle, Radio, and Checkbox to be 32px height with a 4px padding top and bottom. 
The label on the Toggle is also made clickable. 

Note: when text wraps to multiline the clickable area is the following:
```
|----------------|
| [X] Text here  | 
|     text here  |
|     text here  |
|----------------|
```

The main area in question is the area under the checkbox and left of the text being clickable - this is the case for all 3 components. I spoke with Nikola about this and we think that this behaviour is fine.


components/src/Checkbox/Checkbox.js
```diff
@@ -64,13 +64,15 @@ const VisualCheckbox = styled.div`
   }
 `;
 
-const CheckboxWrapper = styled.label`
+const CheckboxClickable = styled.label`
```
clemensp: What do you think about extracting a `ClickableArea` component?

I ask because I noticed that `CheckboxClickable`, `RadioClickable`, and `ToggleClickable` are identical.

clemensp: Or...`InputClickableArea`?
I realized that it's a label, to select the form input.

clemensp (approved):
Left a comment on a possible cleanup.

Looks good. It's nice that `Toggle` now has a `Box` container implementation that mirrors `Radio` and `CheckBox`.

PR #194, opened 2019-02-20
[NDS-478] Add storybook grid background visibility toggle

matthewlyle:
This PR replaces the grid background with a toggleable one that is off by default.

@nikola-nulogy Can you please take a look at the grid? I replaced your new styles with the old storybook styles but the Article story doesn't seem to line up on the grid. It also doesn't line up on the current master version either so I'm not sure if it's supposed to or not.


nikola-nulogy (approved):
Article story looks as I would expect. Not sure what is misaligned.

PR #198, opened 2019-02-21
[NDS-633] Convert weights array to object 

matthewlyle:


components/src/Type/Headings.js
```diff
@@ -8,7 +8,7 @@ Title.defaultProps = {
   m: 0,
   fontSize: 4,
   lineHeight: theme.lineHeights.title,
-  fontWeight: 0,
+  fontWeight: theme.fontWeights.light,
```
clemensp, line 11: Oh man, `light` is so much better than `0`.

clemensp (approved):

PR #201, opened 2019-02-22
[NDS-634] Convert font sizes array to object

matthewlyle:


clemensp (changes requested):
`theme.js` is missing

PR #216, opened 2019-02-28
Nds 668 design a tooltip component

coostenbrug:
Intent: review to merge 
Adds a general tooltip component + stories

Tooltip uses the react-popper library: https://github.com/FezVrasta/react-popper

This tooltip can wrap: Button, Link, Text, Inputs, Box, Flex
This tooltip cannot wrap: Toggle, Checkbox, Radio (causes issues with the current way we have these implemented (they are full width components due to them needing to always be on a newline, perhaps this can be changed in the future)


components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
+import React from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import { Manager, Reference, Popper } from "react-popper";
+import theme from "../theme";
+/* eslint react/destructuring-assignment: 0 */
```
clemensp: 2 questions here:
- Why is there a special rule for this file?
  - We should set eslint rules in this order of preference:
    - Apply globally
    - Disable for specific lines of code using disable/enable
      - https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments
    - Disable for specific file
- Why isn't it closer to where it affects?

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  escFunction(event) {
+    if (event.keyCode === 27) {
+      this.hideTooltip(true);
+    }
+  }
+
+  hideTooltip(skipTimer) {
+    this.clearScheduled();
+    if (!skipTimer) {
+      this.hideTimeout = setTimeout(() => this.setState({ open: false }), this.props.hideDelay);
```
clemensp: I would rename `this.hideTimeout` to `this.hideTimeoutID`.
Currently it sounds more like an action.
Also, the return value is a timeout ID: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+    } else {
+      this.setState({ open: false });
+    }
+  }
+
+  showTooltip() {
+    this.clearScheduled();
+    this.showTimeout = setTimeout(() => this.setState({ open: true }), this.props.showDelay);
```
clemensp: Rename to `showTimeoutID`.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+    default:
+      return ({});
+  }
+};
+
+const TooltipContainer = styled.div({
+  color: tooltipStyles.textColor,
+  display: "flex",
+  flexDirection: "column",
+  fontSize: "14px",
```
clemensp: `small` from the theme.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  backgroundColor: tooltipStyles.backgroundColor,
+  borderRadius: theme.radii.medium,
+  border: `1px solid ${tooltipStyles.borderColor}`,
+  boxShadow: "0 2px 4px rgba(0, 0, 0, 0.18)",
```
clemensp, line 44: Is this going to be a common pattern?
I see the same boxShadow being applied to `MenuItem`.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  padding: theme.space.x1,
+  transition: "opacity 0.3s",
```
clemensp, line 46: I see that this is the same as `MenuItem` as well.
Also, is it intentionally 0.3s as opposed to 0.2s like all the other transitions?

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  ...getTooltipMargin(dataPlacement),
+  ...position,
+  top: open ? 0 : "-9999px",
+  "aria-hidden": !open,
+}));
+
+const getArrowPosition = placement => {
+  const direction = String(placement).split("-")[0];
+  switch (direction) {
+    case "bottom":
```
clemensp: Looks like a refactoring opportunity here.
It feels like a lot of duplication.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  })
+);
+
+class Tooltip extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      open: false,
+    };
+    this.escFunction = this.escFunction.bind(this);
```
clemensp: Rename to `hideTooltipOnEscape`

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+    this.hideTooltip = this.hideTooltip.bind(this);
+    this.showTooltip = this.showTooltip.bind(this);
+  }
+
+  componentWillMount() {
```
clemensp: I'm pretty sure that you should be attaching to the DOM on `componentDidMount`, not on `componentWillMount`.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+
+  getTooltipProps = () => ({
+    onFocus: () => (this.showTooltip()),
+    onBlur: () => (this.hideTooltip()),
+    onMouseEnter: () => (this.showTooltip()),
+    onMouseLeave: () => (this.hideTooltip()),
+    onClick: () => (this.showTooltip()),
+  })
+
+  getElementProps = () => ({
```
clemensp: What does `Element` mean?
I guess I'm wondering where the language comes from.

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+  componentWillUnmount() {
+    document.removeEventListener("keydown", this.escFunction, false);
+  }
+
+  clearScheduled = () => {
+    clearTimeout(this.hideTimeout);
+    clearTimeout(this.showTimeout);
+  };
+
+  getTooltipProps = () => ({
```
clemensp: Would it make sense to make the name more specific?
e.g. `tooltipEventHandlers`

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+
+  render() {
+    return (
+      <Manager>
+        <Reference>
+          {({ ref }) => (
+            <div
+              style={ { display: "inline-flex", minWidth: `${this.props.fullWidth ? "100%" : null}` } }
+              ref={ ref } { ...this.getElementProps() }
+              aria-describedby={ this.props.id }
```
clemensp, line 261: Why does `aria-describedby` need to be here and on the child element?

components/src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,282 @@
...
+},
+({ dataPlacement, open, position }) => ({
+  ...getTooltipMargin(dataPlacement),
+  ...position,
+  top: open ? 0 : "-9999px",
+  "aria-hidden": !open,
+}));
+
+const getArrowPosition = placement => {
+  const direction = String(placement).split("-")[0];
```
clemensp: Is this `location` rather than `direction`?

clemensp (commented):
I think there are some cleanups that can be done before merging.

Willing to pair for the cleanups. :)

components/src/Tooltip/Tooltip.story.js
```diff
@@ -0,0 +1,218 @@
...
+      <Tooltip
+        placement="bottom"
+        tooltip="I am a Tooltip!"
+        id="tooltip1"
+      >
+        <Button> Button </Button>
+      </Tooltip>
+    </Flex>
+  ))
+  .add("with placement", () => (
```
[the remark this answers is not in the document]
coostenbrug: Made the story more clear.

clemensp (commented):
Christiaan: Looks like there was a bad rebase. I see changes from other stories in this branch.

My guess is the branch was rebased after a merge?
Need to fix this. I can help you.

PR #227, opened 2019-03-06
Nds 708 add controllable maxwidth to the tooltip

coostenbrug:
Adds maxWidth to tooltip component

### DIFF
--- a/components/src/Tooltip/Tooltip.story.js
+++ b/components/src/Tooltip/Tooltip.story.js
@@ -33,6 +33,18 @@ storiesOf("Tooltip", module)
       </Tooltip>
     </Flex>
   ))
+  .add("with custom maxWidth", () => (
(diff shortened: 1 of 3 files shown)


components/src/Tooltip/Tooltip.story.js
```diff
@@ -33,6 +33,18 @@ storiesOf("Tooltip", module)
       </Tooltip>
     </Flex>
   ))
+  .add("with custom maxWidth", () => (
+    <Flex justifyContent="center" alignItems="center" height="200px">
```
matthewlyle: Is there a simpler way to get this out from the side of the screen like wrapping it in a Box with padding? Just so it doesn't seem required for implementation or cause questions like "why do we have to justify content for this tooltip?". Not a blocking comment, just a thought.

matthewlyle (approved):

PR #228, opened 2019-03-07
[NDS-697] Add missing component pages

matthewlyle:
This PR adds the following pages to nulogy.design:
- Box
- Flex
- Headings
- Header Validation
- Iconic Button
- Link
- List
- Text
- Textarea

Some have already been merged due to us wanting a filled-out docs site for the engineering meeting, but everything could use a review. Mainly:

- Is everything clear enough for a first iteration of the component docs?
- Are there any inaccuracies? 
- Are any components missing that could be filled out as part of this ticket?


nikola-nulogy (commented):
**Flex**
Can we outline or give a containing flex a colour so it is clearer which element flex is referring to?

nikola-nulogy:
**Headings**
It would be good if the description explained the purpose of the component. (Hierarchy)

nikola-nulogy:
**Header Validation**
- We should add space below the validation just for presentational purposes. The spacing on the form demo is 48px.
- Should link point to the same page, or home, or nothing. I have no idea.

nikola-nulogy:
**Link**
Should the definition include the purpose of the link which is linking to url. This would also remove ambiguity between links and buttons.

nikola-nulogy:
**List**
- Margin-bottom only for presentational purposes should be 24px
- Stoybook link doesn't seem to be working

coostenbrug:
Notices some slight inconsistencies between old and new pages (I like the new better) to do with the checkmark bullet. Example: Uses checkmark bullet on "Content Guidelines" on Button while in HeaderValidation uses checkmark bullet on "Use When" and default bullet on "Content Guidelines"

coostenbrug:
The following "View in storybook" links are broken or not pointed to the right place for me:
 - Headings
 - List 
 - Text
 - Header Validation
For the first 3 the issue is the missing ? in the URL (storybook.nulogy.design/?selectedKind=Component). Not sure why but the HeaderValidation is bringing me to Box but I suspect its due to it having a space in the name in storybook

coostenbrug:
Should we address plural vs singular in the component headings? I see Headings and Buttons are plural because they each contain multiple components, "Icons" perhaps doesnt make sense then and should be Icon? Or maybe everything should always just be singular?

coostenbrug:
In Link: we should mention somewhere that if you do not provide a custom hover color but do provide a custom color, the hover will default to a darkened version of your passed in color

coostenbrug:
Text:

missing disabled prop

coostenbrug (changes requested):

nikola-nulogy (approved):

matthewlyle:
replace checklist with list

clemensp (changes requested):
I see lots of `import { SomeComponent } from @nulogy-components;`

Notice `@nulogy-components`.
Should be `@nulogy/components`.

Other than that, LGTM!

PR #233, opened 2019-03-08
[NDS-713] Replace spacing scale image with code

matthewlyle:
This PR replaces the spacing scale image with code that generates a visual representation of the scale from the `theme.space` object.


docs/src/pages/style/spacing.js
```diff
@@ -1,13 +1,52 @@
...
+  scale: PropTypes.string.isRequired,
+};
+
+const SpacingExamples = props => {
+  const {
+    values,
+  } = props;
+
+  return (
+    Object.keys(values).map(key => {
```
clemensp, line 40: I remembered that there's a better way to do this.

Instead of the guard clause, you can filter for the stuff you want.

```
Object.keys(values)
  .filter(v => v.key !== "none")
  .map(key => <SpacingExample size={ values[key] } scale={ key } />);
```

clemensp (approved):

PR #235, opened 2019-03-12
Nds 676 create a checkboxgroup component

coostenbrug:
Intent: Review to merge

This PR:
 - modifies RadioGroup to use the Legend as its "label"
 - adds HelpText and RequirementText to RadioGroup
 - creates CheckboxGroup that is identical to RadioGroup except there can be multiple defaultValues or checkedValues (uncontrolled vs controlled respectively).

Looking for feedback on:
 - Spacing styling build into the Legend (<legend> cannot be set to "inline" which resulted in me adding RequirementText into the Legend - the drawback of this is the text will be read by a screenreader whenever the legend is read.
 - Should a general "InputGroup" component be pulled out here and then used for the cases of Checkbox and Radio? These are the only two instances for where it is used right now so it may not make sense to pull anything out yet

### DIFF
--- a/components/spec/__snapshots__/Storyshots.spec.js.snap
+++ b/components/spec/__snapshots__/Storyshots.spec.js.snap
@@ -46,6 +46,20 @@ exports[`Storyshots Checkbox Set to error 1`] = `ReactWrapper {}`;
 
 exports[`Storyshots Checkbox Set to required 1`] = `ReactWrapper {}`;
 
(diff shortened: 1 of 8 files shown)


matthewlyle:
1) we should just make sure that screenreaders don't read out required twice. 
2) i don't see any value in abstracting out the inputgroup component, especially since as you said we can always do it later when the value becomes obvious. but others may have a different opinion which i could also get behind. 🤷‍♂️

matthewlyle (approved):

PR #239, opened 2019-03-15
NDS-717-last-navbar-menuitem-gets-pushed-over-when-opened

nikola-nulogy:
Intent: review to merge

### What I Did
Fixed the shifting issue in NavBar


coostenbrug:
Noticed that this works for all MenuDropdowns, but MenuItems are not receiving margin at all (And the last MenuDropdown before a MenuItem does not receive margin.

matthewlyle (approved):

PR #242, opened 2019-03-19
Nds 726 make it obvious when a menu item is a dropdown trigger

nikola-nulogy:
Intent: Review to merge

The goal of this PR was to make it more obvious which menu items expose the drop down and which link to the new page.

### What I did
- Added down arrow to the dropdown triggers
- Added underline to link hover state
- Remove pointer cursor from all buttons

### DIFF
--- a/components/src/Button/Button.js
+++ b/components/src/Button/Button.js
@@ -85,7 +85,7 @@ const Button = styled(BaseButton)(
     verticalAlign: "middle",
     lineHeight: theme.lineHeights.base,
     transition: ".2s",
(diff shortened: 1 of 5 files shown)


matthewlyle:
@nikola-nulogy 	Didn't we decide to show the pointer over all top level links? Could we double check w/ product designers?

nikola-nulogy:
> @nikola-nulogy Didn't we decide to show the pointer over all top level links? Could we double check w/ product designers?

I've put the pointer back. No need to check with designers - I don't think that they had a strong opinion on this.

matthewlyle (approved):

PR #244, opened 2019-03-20
Nds 678 New Field and Form

coostenbrug:
Intent: review to merge

This PR:
 - Removes old Field components
 - Adds new Field component that is just a Box
 - Updates Form to wrap all children with Field (except FormSection, HeaderValidation, Box, and Flex)
 - Updates FormSection to wrap all children with Field (except Box and Flex)

This PR does not:
 - solve the problem of UIDs on the mapping for Form/FormSection children (index is used right now so all children will rerended if the array is changed) This work will be captured in ticket: NDS-738
 - solve the problem of list items for InlineValidation, this PR does remove the example of where it is used, but this issue should be looked into more in this ticket: NDS-739
 - solve all of the console warning issues with DemoForm and DemoPage (this should be checked before merging to master when possible (perhaps we should add something to the pipeline to catch these better? storyshots will catch the issues when you run it locally)


matthewlyle (approved):
Approved but I'd like a +1 from @clemensp re: how we handle the children list as I'm unsure of all the options and if this is the best way of doing it, but it lgtm.

components/src/Form/Form.js
```diff
@@ -4,20 +4,47 @@ import { space } from "styled-system";
...
   children,
   ...props
 }) => (
   <form { ...props }>
     <SectionTitle>{ title }</SectionTitle>
-    { children }
+    { children.map((child, index) => {
+      if (childIsWrappedByField(child.type)) {
+        return (
+          <Field key={ index }>
```
clemensp: Why does this wrapping have to happen here, as opposed to inside each form element?
I ask because `Field` is now nothing but a Box, but being used as a type.

clemensp: If teams add custom components, we can ask them to wrap them in `Field`.
It would be part of `Form`'s public API.

Thoughts?

components/src/Form/FormSection.js
```diff
@@ -1,13 +1,27 @@
...
+  Box,
+  Flex,
+} from "ComponentsRoot";
 import theme from "../theme";
 
 const FormSectionTitle = styled(SubsectionTitle).attrs({
   as: "legend",
 })({});
 
+const doNotWrapElements = [
```
clemensp: Same questions as Form apply here.

clemensp (commented):
I left some questions. Interested in chatting about them before merging.

clemensp (approved):

PR #249, opened 2019-03-22
Remove list reset

matthewlyle:


clemensp (commented):
Why?

It's tricky to review without context: the problem it's solving, the implications of the change, etc

clemensp:
> Why?
> 
> It's tricky to review without context: the problem it's solving, the implications of the change, etc

I just realized that this change may not be ready for review yet!

PR #253, opened 2019-03-27
[NDS-462] Make documentation responsive

matthewlyle:
This PR: 

- Adds responsive values to layout components 
- Adds responsive views for homepage, typography and colour page 
- Turns all hardcoded props tables into propsTables 
- Adds responsive view for propsTables 
- Adds Intro component 
- Adds sticky collapsible sidebar navigation component


nikola-nulogy:
I think that we should center the content in relation to the content area and not the entire page. Something like this.

centered-content

nikola-nulogy:
centered-content

nikola-nulogy (approved):

PR #261, opened 2019-04-01
[NDS-760] Add x5 padding token

matthewlyle:
This PR adds a x5 padding token and fixes a typo on a documentation page.


coostenbrug (changes requested):
Here is the place that the x5 token was applied to in the old branch before it was removed. Can we test that it works here? Other than that looks good! If adding this token works then good to merge
Screen Shot 2019-04-01 at 10 48 56 AM

coostenbrug (approved):

PR #262, opened 2019-04-01
Fix Toggle console errors

matthewlyle:
The toggle consisted of divs wrapped in a p which triggers errors e.g `validateDOMNesting(...): <div> cannot appear as a descendant of <p>.` This PR just replaces the p with a div.


matthewlyle:
Christiaan's approval via [Slack](https://nu.slack.com/archives/DB31FNGAE/p1554134781006300): "Looks good, can you confirm that the screenreader still reads all the label text when you tab into the toggle?" 

Answer: yes, confirmed.

PR #280, opened 2019-04-18
Nds 767 spike react router

coostenbrug:
Intent: Review to merge

This PR:
 - Supports custom render functions in the NavBar
    - API updated to allow for "render", pass in a function
    - custom render is always wrapped with style and onClick
 - Changes spacing styling on Mobile Menu

### DIFF
--- a/components/src/Icon/Icon.js
+++ b/components/src/Icon/Icon.js
@@ -1,9 +1,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
-import { space, color, themeGet } from "styled-system";
+import { space, color } from "styled-system";
 import icons from "../../icons/icons.json";
-import theme from "../theme.js";
+import theme from "../theme";
 
 const iconNames = Object.keys(icons);
 
@@ -21,7 +21,7 @@ const Svg = props => {
       aria-hidden={ title == null }
       width={ size }
       height={ size }
-      fill={ theme.colors[fillColor] ? theme.colors[fillColor] : fillColor}
+      fill={ theme.colors[fillColor] ? theme.colors[fillColor] : fillColor }
       viewBox={ icons[icon].viewBox }
       { ...props }
     >
--- a/components/src/Icon/Icon.js
+++ b/components/src/Icon/Icon.js
@@ -1,8 +1,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 import styled from "styled-components";
-import { space, color } from "styled-system";
+import { space, color, themeGet } from "styled-system";
 import icons from "../../icons/icons.json";
+import theme from "../theme.js";
 
 const iconNames = Object.keys(icons);
 
@@ -20,7 +21,7 @@ const Svg = props => {
       aria-hidden={ title == null }
       width={ size }
       height={ size }
-      fill={ fillColor }
+      fill={ theme.colors[fillColor] ? theme.colors[fillColor] : fillColor}
       viewBox={ icons[icon].viewBox }
       { ...props }
     >
--- a/components/src/NavBar/DesktopMenu.js
+++ b/components/src/NavBar/DesktopMenu.js
@@ -33,13 +33,13 @@ const ApplyMenuLinkStyles = styled.div({
   },
 });
 
-const renderMenuTrigger = (menuItem) => (            
+const renderMenuTrigger = menuItem => (
   <div key={ menuItem.name }>
     <MenuTrigger name={ menuItem.name } menuData={ menuItem.items } />
-</div>
+  </div>
 );
 
-const renderMenuLink = (menuItem) => (
+const renderMenuLink = menuItem => (
   <div key={ menuItem.name }>
     <MenuLink href={ menuItem.href }>
       {menuItem.name}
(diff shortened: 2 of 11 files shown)


components/src/NavBar/DesktopMenu.js
```diff
@@ -6,29 +6,73 @@ import MenuTrigger from "./MenuTrigger";
 import MenuLink from "./MenuLink";
 import theme from "../theme";
 
-const isTrigger = menuItem => (menuItem.items);
+const ApplyMenuLinkStyles = styled.div({
+  "& *": {
```
matthewlyle: i see these in a number of places. does e.g `& *` do anything vs just `*`? i'm pretty sure `& ` renders to nothing. like `li &:hover` would join the parent to be `li:hover`, so `li & *` would join the parent to just be `li *`.

components/src/Icon/Icon.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import styled from "styled-components";
 import { space, color } from "styled-system";
 import icons from "../../icons/icons.json";
+import theme from "../theme";
```
clemensp, line 6: I've been wondering about this one:

What's the point of using `ThemeProvider` if we import themes directly from our default theme file?
I think we should be using `ThemeConsumer` to access the theme.

Probably worth opening a ticket for this one.

components/src/NavBar/DesktopMenu.js
```diff
@@ -6,29 +6,73 @@ import MenuTrigger from "./MenuTrigger";
...
+    return (() => (null));
+  }
+};
 
 const BaseDesktopMenu = ({
   menuData,
   ...props
 }) => (
   <Flex { ...props }>
     {menuData.map(menuItem => {
```
clemensp: Could be shortened to:
`menuData.map(menuItem => getRenderFunction(menuItem)(menuItem))`

which could then be shortened to:
`menuData.map(menuItem => renderMenuItem(menuItem))`

which could be shortened to:
`menuData.map(renderMenuItem)`

Thoughts?

components/src/NavBar/DesktopMenu.js
```diff
@@ -6,29 +6,73 @@ import MenuTrigger from "./MenuTrigger";
...
+
+const getRenderFunction = menuItem => {
+  if (menuItem.items) {
+    return renderMenuTrigger;
+  } else if (menuItem.href) {
+    return renderMenuLink;
+  } else if (menuItem.render) {
+    return renderCustom;
+  } else {
+    return (() => (null));
```
clemensp, line 64: I'm not sure if it's safe to include a `renderNothing` fallback here.
If there's nothing, it's an error, right?

I would consider raising an exception here:
`throw new Error("nothing to render")`?

I'm actually not sure if it's a good practice to throw errors inside render functions, so I would do a quick search for that. I did a 1 minute search, and didn't find much.

components/src/NavBar/MobileMenu.js
```diff
@@ -12,42 +12,128 @@ import MenuLink from "./MenuLink";
...
+  textDecoration: "none",
+  lineHeight: theme.lineHeights.base,
+  transition: ".2s",
+  fontSize: theme.fontSizes.medium,
+  "& *": {
+    display: "block",
+    color: theme.colors.white,
+    textDecoration: "none",
+    padding: `${theme.space.x1} ${theme.space.x2}`,
+    paddingLeft: `${(24 * layer) + 24}px`,
```
clemensp: I would extract the padding logic into a function. I see it appear several times.

components/src/NavBar/MobileMenu.js
```diff
@@ -88,48 +179,46 @@ const Menu = styled.ul(() => (
...
+    "&:hover, &:focus": {
+      outline: "none",
+      color: theme.colors.lightBlue,
+      backgroundColor: theme.colors.black,
+      cursor: "pointer",
+    },
+  }
+);
+
 const MobileMenuBase = ({
```
clemensp: Nice, looked at the file, and the function reads well.

clemensp (changes requested):
Looks nice.

Some comments left.

Also, make sure you add the changes to `CHANGELOG.md` once you rebase.

components/src/NavBar/DesktopMenu.js
coostenbrug: Will make these changes, & will render to nothing while not including & will render to a space with basic nesting (hence why it is needed in the li &:hover example but not in these cases.

components/src/Icon/Icon.js
coostenbrug: This makes sense, making a ticket:

https://nulogy-go.atlassian.net/browse/NDS-817

components/src/NavBar/DesktopMenu.js
coostenbrug: I cleaned up this instance in DesktopMenu and in MenuTrigger and SubMenuTrigger. The mapping in MobileMenu.js is a bit more complex on how the render function is determined and what is passed to it so I left that one a bit more verbose to keep it readable.

components/src/NavBar/DesktopMenu.js
coostenbrug: Ideally this is handled with PropTypes. I've made a function to validate that menuData is correct in structure and content and throws errors if not.

PR #287, opened 2019-04-29
Nds 829 navbarsearch input unstyled in safari

nikola-nulogy:
Intent: Review to merge

What I did:
- forced webkit browsers to treat input type search as textfield input
- Added 0 margin to input to fix the spacing bug in safari - Select filed was followed by some margin from some reason.

### DIFF
--- a/components/src/Input/Input.js
+++ b/components/src/Input/Input.js
@@ -42,6 +42,7 @@ const StyledInput = styled.input(
     fontSize: theme.fontSizes.medium,
     fontFamily: theme.fonts.base,
     lineHeight: theme.lineHeights.base,
+    margin: theme.space.none,
     "&:focus": {
       outline: "none",
       color: theme.colors.black,
--- a/components/src/Input/Input.js
+++ b/components/src/Input/Input.js
@@ -52,12 +52,6 @@ const StyledInput = styled.input(
     "::placeholder": {
       color: transparentize(0.4, theme.colors.black),
     },
-    "&[type='search']": {
-      "-webkit-appearance": "textfield",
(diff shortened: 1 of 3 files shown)


matthewlyle (approved):

components/src/Input/Input.js
```diff
@@ -42,6 +42,7 @@ const StyledInput = styled.input(
     fontSize: theme.fontSizes.medium,
     fontFamily: theme.fonts.base,
     lineHeight: theme.lineHeights.base,
+    margin: theme.space.none,
```
coostenbrug, line 44: Setting margin here after space is above prevents the use of the "m" space props. One solution would be to move space below these styles

components/src/Input/Input.js
```diff
@@ -51,6 +52,12 @@ const StyledInput = styled.input(
     "::placeholder": {
       color: transparentize(0.4, theme.colors.black),
     },
+    "&[type='search']": {
```
coostenbrug: Could this be put directly on the NavBarSearch component instead of on the Input? Keeps the input component a bit cleaner and keeps relevant stylings directly on the component we want to apply it to.

coostenbrug (changes requested):

components/src/Input/Input.js
nikola-nulogy: I was debating where this should go and originally had it in navbar search component but then it seemed like more of a reset thing that should be pushed level up. But I agree it would be less confusing if it was in navbar search component so I'll move it there.

coostenbrug (approved):

PR #294, opened 2019-05-02
Resets buttons margin and fixes Safari bug

nikola-nulogy:
Intent: Review to merge

What I did
- Reset button margin

### DIFF
--- a/components/src/Button/Button.js
+++ b/components/src/Button/Button.js
@@ -74,7 +74,6 @@ BaseButton.defaultProps = {
 
 
 const Button = styled(BaseButton)(
-  space,
   size,
   ({ disabled, fullWidth }) => ({
     display: "inline-flex",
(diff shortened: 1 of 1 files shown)


components/src/Button/Button.js
```diff
@@ -91,6 +91,7 @@ const Button = styled(BaseButton)(
     border: `1px solid ${theme.colors.darkBlue}`,
     borderRadius: theme.radii.medium,
     width: fullWidth ? "100%" : "auto",
+    margin: theme.space.none,
```
coostenbrug, line 93: Since space is above margin this will override any props passed in using styled-system margin. We probably want to allow margin changes for buttons as I can see some use cases for it. The easy fix would be to move space below the styles object here

coostenbrug (changes requested):

coostenbrug (approved):

PR #303, opened 2019-05-17
Nds 869 expose an api for list of errors for inline form errors

coostenbrug:
Intent: Review to merge


components/src/Checkbox/CheckboxGroup.js
```diff
@@ -9,39 +9,44 @@ import { Fieldset } from "../Form";
...
+      {getCheckboxButtons(otherProps)}
+      <InlineValidation mt="x1" errorMessage={errorMessage} errorList={errorList} />
+    </Fieldset>
+  );
+};
 
 BaseCheckboxGroup.propTypes = {
-  error: PropTypes.string,
+  errorMessage: PropTypes.string,
+  errorList: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
```
clemensp: Why support arrays and strings for `errorList`?

components/src/Select/Select.js
```diff
@@ -152,7 +153,7 @@ const Select = ({
         highlightedIndex
       }) => (
         <div style={{ position: "relative" }}>
-          <SelectBox {...getToggleButtonProps({ disabled, error, isOpen })}>
+          <SelectBox {...getToggleButtonProps({ disabled, error: !!(errorMessage || errorList), isOpen })}>
```
clemensp: Looks like `!!(errorMessage || errorList)` appears in a *lot* of places.
Definitely an opportunity for cleanup.

coostenbrug: Did some cleanup here

components/src/Checkbox/CheckboxGroup.js
coostenbrug: Removed the support for the string case, I agree it made more sense to enforce it to always be an array

clemensp (approved):

PR #311, opened 2019-06-05
[NDS-921] Add light colours from alert to design system 

matthewlyle:
This PR updates the design system to include the new light green, red and yellow.


nikola-nulogy (changes requested):
Light yellow doesn't seem to show up on doc site.

PR #315, opened 2019-06-11
[NDS-922] Align validation styles to Alert

matthewlyle:
This PR makes HeaderValidation use the new Alert component instead of a custom style. 

We (Nikola and I) decided not to update any inline validation styles as they already shared the icon and red colour with the Alerts and adding the background made everything a little intense.


coostenbrug (approved):
Looks good! Just fixed one console warning by passing the props spread through to the wrapping box. Good to merge once it passes CI

PR #328, opened 2019-06-18
[NDS-946] Differentiate focus and hover more

matthewlyle:
This PR updates the focus shadow token to have greater depth and less translucency, and removes the darkened background on buttons that are focused but not hovered.


coostenbrug (approved):
Changes look good!

Additional request: 
Can you check out the MenuDropdown and the Navbar (both mobile and desktop sizes) to confirm those buttons/links are styled as intended? They still use the old background color change on focus, but the context is very different so it may be okay there?

PR #339, opened 2019-07-09
Fix docs site accessibility issues

matthewlyle:
We had some accessibility issues on our docs site causing: 
* 78/100 in Google lighthouse
* 59 violations in aXe 

The issues were caused by lack of contrast in the nav and highlight styles and improper usage of `<li>`s.


matthewlyle:
Found an issue: the close button in the mobile nav is darkblue. We need to update iconicbutton to support different colours, or just use a regular link with an icon inside of it with appropriate colours.

PR #342, opened 2019-07-11
Use the correct cursor for DropdownButton when disabled

JonErikDSuero:


components/src/DropdownMenu/DropdownButton.js
```diff
@@ -6,7 +6,7 @@ import theme from "../theme";
 const DropdownButton = styled.button(props => ({
   display: "block",
   width: "100%",
-  cursor: "pointer",
+  cursor: props.disabled ? "arrow" : "pointer",
```
JonErikDSuero, line 9: Do I need to create a new example in Storybook?

matthewlyle:
i don't think "arrow" is the right cursor. i believe it's "default". https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

JonErikDSuero:
@matthewlyle: I'm copying https://github.com/nulogy/design-system/blob/43b2cbc5aee7238fd9d99fbb86ab869ef0856438/components/src/Button/Button.js#L74

Should I update that code also?

matthewlyle:
interesting find. it looks like that code's actually not doing anything and can be removed. i think the "disabled" tag on the button itself is what's causing the cursor to look like that.

matthewlyle:
give me a second to look into this.

matthewlyle:
so we don't actually need any cursor css styling on disabled buttons. what's missing is the `disabled` prop on the dropdownMenu component, to pass disabled down to the actual button.

JonErikDSuero:
will be fixed by another solution

PR #345, opened 2019-07-17
[NDS-989] Add Select improvements 

matthewlyle:
This PR switches our Select from using Downshift to using React-Select and adds some requested functionality. The api should be the same and the component shouldn't require any migration other than the fact that the field's value is submitted instead of its label (this is apparently what was always desired).


matthewlyle:
**Current status:** 

✅ 	Cypress tests (although we're referencing the text inside the singlevalue class, rather than an actual value, since react-select doesn't seem to store that in the dom)
🚫 Enzyme tests

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Select updates
+  - Now uses React-Select instead of Downshift
+  - [**Breaking Change**] When submitting a form, Select's value is submitted instead of label
+  - Default maxHeight changed to 246px to provide better scrolling affordance and ScrollIndicators removed
+  - Added proper focus token
+  - Added autocomplete
```
clemensp, line 26: Consider linking to the story in storybook

components/src/Select/Select.story.js
```diff
@@ -142,10 +161,10 @@ storiesOf("Select", module)
       helpText="Additional information about input"
     />
   ))
-  .add("With smaller maxHeight", () => (
+  .add("with smaller maxHeight", () => (
     <Select
       initialIsOpen
-      maxHeight="128px"
+      maxHeight="132px"
```
clemensp, line 167: What's this change for?

components/src/Select/Select.story.js
```diff
@@ -33,23 +33,33 @@ class SelectWithState extends React.Component {
...
+      <>
+        <Select onChange={this.handleChange} value={selectedValue} options={options} {...this.props} />
+        <Button onClick={this.clearSelection}>Clear selection</Button>
+      </>
+    );
   }
 }
 
 storiesOf("Select", module)
   .add("Select", () => (
```
clemensp, line 59: clicking the label doesn't open the dropdown.

components/src/Select/Select.story.js
```diff
@@ -67,10 +77,10 @@ storiesOf("Select", module)
...
   ))
-  .add("Select as a controlled component", () => (
+  .add("as a controlled component", () => (
     <SelectWithState placeholder="Please select inventory status" options={options} labelText="Inventory status" />
   ))
-  .add("Set to disabled", () => (
+  .add("set to disabled", () => (
     <Select placeholder="Please select inventory status" options={options} disabled labelText="Inventory status" />
   ))
   .add("with error message", () => (
```
clemensp, line 86: The error message box, when open, has a red border with blue drop-shadow. Is this intended?

components/src/Select/Select.story.js
matthewlyle: To show a better example of scrolling affordance when using maxHeight. It now cuts off in the middle of a word, instead of near the top, making it very obvious that there is scrollable content in the box.

components/src/Select/Select.story.js
matthewlyle: It was intended but will now be changed :smile:

matthewlyle:
Tests are now working, but we're having an issue building the docs site. (You can see the full error message in [travis](https://travis-ci.org/nulogy/design-system/builds/566443418?utm_source=github_status&utm_medium=notification)). I'm not sure what the problem is... but it can be replicated locally if you try to run `yarn serve` in the docs package, which will do a production build. (`yarn start`, the dev server, works fine). 

The error is `error #95313 ReferenceError: document is not defined`. Looking that up brings me to a helpful looking [Gatsby page](https://www.gatsbyjs.org/docs/debugging-html-builds) but I'm not sure how to troubleshoot which of these issues it is and how to fix it.

PR #347, opened 2019-07-22
Modal component

coostenbrug:


matthewlyle (commented):
design and docs look good to me 👍

components/src/Modal/Modal.js
```diff
@@ -0,0 +1,308 @@
...
+      </Button>
+    ));
+  }
+
+  getModalButtons() {
+    const { buttonAlignment } = this.props;
+
+    if (buttonAlignment === "spaced") {
+      return (
+        <React.Fragment>
```
clemensp: `<>`

components/src/Modal/Modal.js
```diff
@@ -0,0 +1,308 @@
...
+                  {title}
+                </SectionTitle>
+              ) : (
+                <div style={{ height: theme.space.x4 }} />
+              )}
+              {onRequestClose && <ModalCloseButton onClick={onRequestClose} />}
+            </ModalHeader>
+          )}
+          <ModalContent>
+            <InnerModalContent>{children}</InnerModalContent>
```
clemensp: `SomeComponent` and `InnerSomeComponent` is a design smell.

Why do we need `ModalContent` and `InnerModalContent`?
Is there a way to handle `InnerModalContent` inside `ModalContent`?

components/src/Modal/Modal.story.js
```diff
@@ -0,0 +1,140 @@
...
+            <Input name="name" id="name" labelText="Name" />
+            <Input type="number" name="age" id="age" labelText="Age" />
+          </Form>
+        </Modal>
+      </div>
+    );
+  }
+}
+
+storiesOf("Modal", module)
```
clemensp, line 64: Unused param `module`

docs/src/pages/components/navbar.js
```diff
@@ -14,6 +14,7 @@ import {
   ListItem,
   Text
 } from "@nulogy/components";
+import { theme } from "@nulogy/components/src";
```
clemensp: `theme` is exported by `@nulogy/components`, so you should be able to add it to the list above.

clemensp (approved):
LGTM.
Left a few comments to consider. :)

components/src/Modal/Modal.js
coostenbrug: Sounds good, Ive changed all instances of React.Fragment to use this syntax to be consistent

components/src/Modal/Modal.story.js
coostenbrug: module is required for storybook to have hot-refreshing: https://github.com/storybookjs/storybook/issues/2022

PR #348, opened 2019-07-22
[RFC, NO MERGE] Typescript For Components

ecbrodie:
**THIS PR CONVERTS ALL OF NDS COMPONENTS INTO TYPESCRIPT!!!**

This is likely not ready for merge yet, there are many `tsc` errors that still need to be fixed and the one day I had to make this PR was not enough to deal with them. However, Storybook is working and Webpack builds will also likely work too because utilizing Babel enables NDS to have partially functional Typescript.

My methodology was to convert all files at once. Perhaps it is more desirable to follow an approach where files are converted one at a time.

Regardless, I'd love to get feedback on my work. If the NDS maintenance team is interested in continuing forth with supporting Typescript, then accepting this PR when it is more polished is advisable.


ecbrodie:
I'm closing this out based on IRL conversation I had with @clemensp . We agreed that the first step towards Typescript adoption is a smaller PR that just adds bare-bones infrastructure, such as `tsconfig.json` and Babel upgrades.

PR #355, opened 2019-07-31
Remove label attribute on buttons

coostenbrug:
This PR fixes error where IconicButton and CloseButton had the label attribute instead of just the aria-label attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - [Button](https://nulogy.design/components/buttons/) now uses the proper border colour (`blue`)
+- [IconicButton](https://nulogy.design/components/iconic-button/) now uses the `aria-label` attribute instead of the `label` attribute
 
 - NavBar bug fixes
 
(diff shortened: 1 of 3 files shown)


clemensp:
> This PR fixes error where IconicButton and CloseButton had the label attribute instead of just the aria-label attribute.

This tells me what changed, but not what was fixed.
e.g. Was something behaving incorrectly?

coostenbrug:
> > This PR fixes error where IconicButton and CloseButton had the label attribute instead of just the aria-label attribute.
> 
> This tells me what changed, but not what was fixed.
> e.g. Was something behaving incorrectly?

Sorry, should have made that clear in initial comment. the label attribute was added mistakenly to the button component (it wont do anything for screen-readers). Instead the aria-label attribute was the one that was intended to be attached as screen readers will read that

clemensp (approved):

clemensp:
> > > This PR fixes error where IconicButton and CloseButton had the label attribute instead of just the aria-label attribute.
> > 
> > 
> > This tells me what changed, but not what was fixed.
> > e.g. Was something behaving incorrectly?
> 
> Sorry, should have made that clear in initial comment. the label attribute was added mistakenly to the button component (it wont do anything for screen-readers). Instead the aria-label attribute was the one that was intended to be attached as screen readers will read that

No need to be sorry. I just needed more info to be able to review. =)
Thanks for the fix! :)

PR #366, opened 2019-08-12
Nds 1019 tab component

coostenbrug:


components/src/Tabs/TabFocusManager.js
```diff
@@ -0,0 +1,96 @@
...
+      default:
+        break;
+    }
+  }
+
+  focusNextTab() {
+    const { tabRefs } = this.props;
+
+    this.setState(prevState => ({
+      focusedIndex: prevState.focusedIndex === tabRefs.length - 1 ? 0 : prevState.focusedIndex + 1
```
clemensp: `(prevState.focusedIndex + 1) % tabRefs.length` also works if you want a cool solution. :D

components/src/Tabs/TabFocusManager.js
```diff
@@ -0,0 +1,96 @@
...
+    }));
+  }
+
+  focusPreviousTab() {
+    const { tabRefs } = this.props;
+
+    this.setState(prevState => ({
+      focusedIndex: prevState.focusedIndex === 0 ? tabRefs.length - 1 : prevState.focusedIndex - 1
```
clemensp: So sad. In Ruby, the same line as above would work.
But in javascript. `-1 % 5` returns `-1`.
In ruby, `-1 % 5` returns `4`.

`(prevState.focusedIndex - 1 + tabRefs.length) % tabRefs.length` would work, but it's not as nice. :(

components/src/Tabs/TabFocusManager.js
```diff
@@ -0,0 +1,96 @@
...
+          setFocusToTab: this.setFocusToTab
+        })}
+      </>
+    );
+  }
+}
+
+TabFocusManager.propTypes = {
+  children: PropTypes.func.isRequired,
+  tabRefs: PropTypes.arrayOf(PropTypes.shape({ current: PropTypes.instanceOf(Element) }))
```
clemensp: Is `PropTypes.instanceOf(Element)` different from `PropTypes.element`?

components/src/Tabs/TabScrollIndicators.js
```diff
@@ -0,0 +1,155 @@
...
+    for (let i = 0; i < tabRefs.length; i += 1) {
+      scrollLeftSum += tabRefs[i].offsetWidth;
+      if (leftMarker <= scrollLeftSum) {
+        return i;
+      }
+    }
+    return null;
+  }
+
+  handleScroll() {
```
clemensp, line 94: Does this trigger when the scrolling ends? Or several times during a scroll?

components/src/Tabs/Tabs.story.js
```diff
@@ -0,0 +1,178 @@
...
+      <Tab label="Tab 6">Tab 6 Content</Tab>
+      <Tab label="Tab 7">Tab 7 Content</Tab>
+      <Tab label="Tab 8">Tab 8 Content</Tab>
+      <Tab label="Tab 9">Tab 9 Content</Tab>
+      <Tab label="Tab 10">Tab 10 Content</Tab>
+      <Tab label="Tab 11">Tab 11 Content</Tab>
+      <Tab label="Tab 12">Tab 12 Content</Tab>
+    </Tabs>
+  ))
+  .add("TESTING tabs", () => (
```
clemensp: What's "TESTING"?

clemensp (approved):
LGTM overall.

Feel free to address my comments if they make sense. :)

components/src/Tabs/Tabs.story.js
coostenbrug: Removed these stories and attached them to the testing ticket, we can resuse them when writing Cypress tests later

components/src/Tabs/TabScrollIndicators.js
coostenbrug: Instead of going with a debounce or a throttle (as it would cause the ScrollIndicator on each side not appear/disappear when hitting the end or start of the container) I went with the approach of removing the scrollLeft state altogether. Instead I added contentHiddenLeft and contentHiddenRight boolean states which update only when they change which causes the ScrollIndicators to re-render. onScroll will now only have to run the if statement which is much less taxing than constantly setting state causing re-renders

PR #393, opened 2019-10-02
Make links always black in alerts and add examples

marboro92:
- Overrides the color for all links within <Alert> to be black (even if a color is set on the <Link>)
- added an example of alerts with links to both storybook and the docs


clemensp (approved):
Looks good!

PR #406, opened 2019-10-16
Status indicator

nikola-nulogy:
What I did

- Added StatusIndicator component
- Added doc site article

### DIFF
--- a/components/src/StatusIndicator/StatusIndicator.js
+++ b/components/src/StatusIndicator/StatusIndicator.js
@@ -76,7 +76,7 @@ const StatusIndicator = styled.p(
 );
 
 StatusIndicator.propTypes = {
-  type: PropTypes.oneOf(["danger", "informative", "success", "warning", "quiet"]),
+  type: PropTypes.oneOf(["neutral", "danger", "informative", "success", "warning", "quiet"]),
   ...space.PropTypes
 };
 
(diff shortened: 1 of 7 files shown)


components/src/StatusIndicator/StatusIndicator.js
```diff
@@ -0,0 +1,88 @@
...
+  warning: {
+    borderColor: theme.colors.yellow,
+    backgroundColor: theme.colors.yellow,
+    color: theme.colors.darkGrey
+  }
+};
+
+const getStatusIndicatorColours = props => {
+  if (props.type === "neutral") {
+    return StatusIndicatorColours.neutral;
```
marboro92: could do StatusIndicatorColours[props.type] instead of the if statements for each type here

components/src/StatusIndicator/StatusIndicator.js
```diff
@@ -0,0 +1,88 @@
...
+    fontWeight: theme.fontWeights.bold,
+    textTransform: "uppercase",
+    letterSpacing: ".05em",
+    padding: `0 ${theme.space.x1}`,
+    borderRadius: theme.space.x1,
+    position: "relative",
+    top: "-2px"
+  },
+  space,
+  props => getStatusIndicatorColours(props)
```
marboro92: it would be better to pass in only what's needed into the function here. i.e. pass only type to getStatusIndicatorColours rather than all props

marboro92 (approved):

PR #407, opened 2019-10-16
NDS-1142 adds example of dropdown as a custom cell

marboro92:
- added a story in story book for using dropdowns in tables
- Just referenced storybook for more examples of using customCellRenderer in the docs to avoid adding too much there for each use case


docs/src/pages/components/table.js
```diff
@@ -142,14 +143,13 @@ const rows = [{ c1: "row 1 cell 1", c2: "r1c2", c3: "2019-09-21" }, { c1: "r2c1"
       <SectionTitle>With a custom component</SectionTitle>
       <Text>
         Providing a cellRenderer function inside the column data will allow
-        display of arbitrary cell content.
+        display of arbitrary cell content. See storybook for other examples of
```
matthewlyle: We should probably link to the storybook any time we mention it.

matthewlyle (approved):
Approved with one non-blocking suggestion.

PR #413, opened 2019-10-18
Support array value props for multiselect

ecbrodie:
This PR was created in coordination with @Fitzsimmons 

The props `defaultValue` and `value` in the `<Select>` component currently only support single elements, as the internal `getValue()` function would only find one option object matching the value before providing that option to the React-Select component. This means that being able pre-populate this component as a multi-select with > 1 value (either as a controlled component with `value` or an uncontrolled component with `defaultValue`) was not supported.

This PR addresses that issue by adding support to these props to optionally be an array. This PR also renames some variables and functions to make the code more clear.

## TODO

* [ ] I would like to add documentation for these new props. The problem is that `defaultValue` and `value` are documented in `docs/src/shared/inputProps.js`, so making a change for just `<Select>` would incorrectly affect all other NDS components that use these props. What is the correct way to document this?
* [ ] Does a storybook example for multiselect values need to be added?
* [ ] Does such an example also need to be added to the documentation page?

### DIFF
--- a/components/src/Select/Select.js
+++ b/components/src/Select/Select.js
@@ -6,10 +6,18 @@ import { MaybeFieldLabel } from "../FieldLabel";
 import { InlineValidation } from "../Validation";
 import customStyles from "./customReactSelectStyles";
 
-const getValue = (opts, val) => {
-  if (val === "") return val;
+const getOption = (options, value) => {
+  if (value == null || value === "") return value;
 
-  return opts.find(o => o.value === val);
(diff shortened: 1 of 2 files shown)


marboro92 (approved):

marboro92:
Thanks for updating this! It looks great to me. 
As far as the TODO items, go.. I am going to create a story for us to revisit the documentation/storybook for this and that will get updated soon.

PR #416, opened 2019-10-24
NDS-1078 Pagination component

matthewlyle:


components/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,139 @@
...
+  color: props.currentPage ? theme.colors.whiteGrey : theme.colors.black
+}));
+
+const SEPERATOR = "...";
+
+const getPageItemstoDisplay = (totalPages, currentPage) => {
+  const pages = Array.from({ length: totalPages }, (v, k) => k + 1);
+  if (totalPages <= 5) return pages;
+  if (currentPage === 1) return [...pages.slice(currentPage - 1, 2), SEPERATOR, totalPages];
+  if (currentPage === totalPages) return [1, SEPERATOR, ...pages.slice(totalPages - 2, totalPages)];
```
marboro92, line 85: Just noticed that we might have to fix this for  currentPage <= 3..
since we should have 
[1, 2*, 3, ..., 10]
[1, 2, 3*, 4,..., 10]
but currently i think we would get 
[1, .. 1, 2*. 3,... 10]
and same for 3 numbers from the last page

marboro92 (approved):

PR #418, opened 2019-10-25
NDS-1161 client side pagination

marboro92:
- Adds client side pagination, which is determined by a rowsPerPage prop
- Supports client-side pagination in combination with selectable rows, and keeps track of the state of the checkboxes per page 
- Adds functional and unit tests for core functionality of pagination on its own and within the table
- cypress e2e tests created and cleaned up to test the paginated table and selectable row functionality


matthewlyle:
Everything looks good to me except we need to add 16px of space between the table and the pagination. 
Screen Shot 2019-10-25 at 10 21 41 AM

matthewlyle (approved):

PR #421, opened 2019-10-25
NDS-1148 document using server side pagination

marboro92:
Documents an example of using the Table and Pagination together with data fetched from jsonplaceholder

### DIFF
--- a/components/spec/Storyshots.spec.js
+++ b/components/spec/Storyshots.spec.js
@@ -3,5 +3,6 @@ import { mount } from "enzyme";
 import "jest-styled-components";
 
 initStoryshots({
-  renderer: mount
+  renderer: mount,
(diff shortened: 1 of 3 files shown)


components/spec/Storyshots.spec.js
```diff
@@ -3,5 +3,6 @@ import { mount } from "enzyme";
 import "jest-styled-components";
 
 initStoryshots({
-  renderer: mount
+  renderer: mount,
+  storyNameRegex: /^((?!.*?SkipStoryshot).)*$/
```
marboro92, line 7: Apparently looking for a string within a story name is the only way to skip a storyshot for a particular story.. 

It needs to be skipped since it relies on using fetch for the purposes of the example.

matthewlyle (approved):

PR #427, opened 2019-10-31
NDS-1182 support table footers

marboro92:


components/src/Table/BaseTable.js
```diff
@@ -4,44 +4,53 @@ import PropTypes from "prop-types";
 
 import TableHead from "./TableHead";
 import TableBody from "./TableBody";
+import TableFoot from "./TableFoot";
+
+const columnsPropType = PropTypes.arrayOf(
```
matthewlyle: could be imported from Table.types

matthewlyle (approved):

PR #429, opened 2019-11-01
Nds 1172 modal cuts off select options

nikola-nulogy:
### What I did:

- Removed `overflow: hidden` from Modal wrapper and adjusted spacing
- Added stories with select (select, select with scrolling, select without actions ...)

### Solution
- Make elements that potentially bleed out of the screen (select options, date picker ...) small. Select should be 96px heigh.
- Use space after to provide more room (select options in this case)

With the proposed solution `overflow: auto` is preserved which allowed scrolling. The alternative of allowing content to bleed out with `overflow: visible` was considered but that broke scrolling. The chosen approach was considered superior because in the case when the content overflows modal it can also overflow the viewport and that content would not be accessible.

### Other possibilities

1. Providing specific styles and behaviors for the case when modal scrolls and when it doesn't
2. Removing max-height from Modal so there is no scrolling and we can use only `overflow: visible`. This would result in a modal header and footer going out of viewport when scrolled.

### DIFF
--- a/components/src/Modal/Modal.js
+++ b/components/src/Modal/Modal.js
@@ -11,7 +11,6 @@ import { PreventBodyElementScrolling } from "../utils";
 
 const ModalContent = styled.div({
   marginTop: "-64px",
-
   marginBottom: "-72px",
   overflow: "auto",
   paddingTop: "88px",
--- a/components/src/Modal/Modal.js
+++ b/components/src/Modal/Modal.js
@@ -77,7 +77,6 @@ const StyledReactModal = styled(ReactModal)(
     left: 0,
     right: 0,
     bottom: 0,
-    overflow: "hidden",
     backgroundColor: theme.colors.white,
     borderRadius: theme.radii.medium,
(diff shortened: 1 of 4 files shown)


components/src/Modal/Modal.js
```diff
@@ -11,10 +11,11 @@ import { PreventBodyElementScrolling } from "../utils";
 
 const ModalContent = styled.div({
   marginTop: "-64px",
-  marginBottom: "-80px",
+
```
marboro92: can you remove the space here? 
Also I'm not quite clear on why these margins and paddings are changing, are they just style tweaks or specific changes for the select?

marboro92 (approved):

components/src/Modal/Modal.js
nikola-nulogy: The new values allow more space for the select options. Also, I corrected some incorrect spacing where ModalContent was 8 or 16px larger than its parent - StyledReactModal.

PR #431, opened 2019-11-04
Required prop on input field doesn't work

nikola-nulogy:
### What I did
- Added required prop to Input and Textarea
- Created 'set to required' stories for input and textarea


matthewlyle:
Can we add a form with a submit button to the required storybook examples? That way it can test that it works (e.g if required tag isn't working, submitting the form will succceed).  I don't mind doing the work for this.

matthewlyle (approved):

PR #435, opened 2019-11-07
NDS-1190 Changes Buttons to QuietButtons in form examples

matthewlyle:


marboro92:
I noticed there's a storybook example  (modal--styled-as-a-form) with a form and cancel button using Button rather than QuietButton and the same example in the docs for modal, should those get updated too?

PR #442, opened 2019-11-11
add custom column widths to table headers

marboro92:
Width was removed from the TableCells in some earlier code clean up, but width was missing in the table headers which caused the custom widths in table to break.


matthewlyle (approved):
nice catch

PR #469, opened 2019-11-25
[NDS-1223] stop footer cells from being affected by column cellrenderer

marboro92:
## Description

Table footers will no longer use the cellRenderer or cellFormatter defined on columns and the headers for the table footers have been moved to the left with the appropriate colSpan depending on whether there are control columns added (selectable column or expandable).

## Changes include

- [ ] Breaking change (change that is not backwards-compatible and/or changes current 
functionality)
- [ ] Documentation only

## Checklist

- [x] Changelog updated

## Before Merging

- [x] Tested storybook deployment preview
- [x] Tested docs deployment preview

### DIFF
--- a/components/src/Table/TableFoot.js
+++ b/components/src/Table/TableFoot.js
@@ -15,19 +15,32 @@ const StyledFooterRow = styled.tr({
 const renderRows = (rows, columns, keyField, loading) =>
   rows.map(row => <TableFooterRow row={row} columns={columns} key={row[keyField]} loading={loading} />);
 
-const TableFooterRow = ({ row, columns, loading }) => (
-  <StyledFooterRow>
-    {columns.map((column, index) =>
-      index === 0 ? (
-        <StyledTh key={column.dataKey} scope="row">
-          {row[column.dataKey]}
-        </StyledTh>
-      ) : (
-        !loading && <TableCell key={column.dataKey} row={row} column={column} cellData={row[column.dataKey]} />
-      )
-    )}
-  </StyledFooterRow>
-);
+const TableFooterRow = ({ row, columns, loading }) => {
+  console.log(columns);
+  const columnsWithoutControls = columns.filter(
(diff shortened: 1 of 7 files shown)


components/src/Table/TableFoot.js
```diff
@@ -15,19 +15,37 @@ const StyledFooterRow = styled.tr({
...
+  // const renderEmptyCells = () =>
+  //   Array(numberOfControlColumns)
+  //     .fill()
+  //     .map(() => <td />);
+  return (
+    <StyledFooterRow>
+      {/* {renderEmptyCells()} */}
+      {columnsWithoutControls.map((column, index) =>
+        index === 0 ? (
+          <StyledTh key={column.dataKey} scope="row" colSpan={numberOfControlColumns + 1}>
```
marboro92, line 27: The commented code here is for adding in EmptyCells for the control columns.
I commented it out because to me it looked sort of strange to have the headers so far into the table.. So I added another option here using colSpan that moved the Headers to the very left.

Here are some screenshots of the two options:

COLSPAN: 
Screen Shot 2019-11-25 at 4 56 04 PM

EMPTY TDs: 
Screen Shot 2019-11-25 at 4 58 05 PM

Let me know what you think.

matthewlyle: I think I prefer the empty TDs because it puts the label closer to its associated content. @nikola-nulogy thoughts?

nikola-nulogy: I think I slightly prefer colspan. I like how it looks different from the data portion of the table reducing the chance of confusing it for the data. Empty td does put it closer to the content but that works only if the following column has a total value which may or may not be the case.

matthewlyle: Ok let's go with colspan for now then.

matthewlyle (approved):

PR #470, opened 2019-11-26
[NDS-1221] after selecting a row with a keyboard the focus is reset to the select all checkbox

marboro92:
## Description

Using the HOC pattern we had before caused the upper level component's ref to get get lost between rerenders when both withSelectableColumn and withExpandableColumn were combined, and thus caused the focus to be lost whenever we changed the row data by expanding or selecting an item.

The only way I could get around this was to change the HOC's for utils (addSelectableControl, addExpandableControl) that only transform the necessary parts of the props rather than returning a new component.

## Changes include

- [ ] Documentation only

## Checklist

- [x] Changelog updated
    _Enter key functionality for selecting rows is still missing here, since its also missing in the checkbox itself, which I'll create a ticket for_

## Before Merging

- [x] Tested storybook deployment preview


matthewlyle:
>   _Enter key functionality for selecting rows is still missing here, since its also missing in the checkbox itself, which I'll create a ticket for_

The expected key for selecting a checkbox is space, not enter. ([source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#description), also see html checkbox default behaviour)

marboro92:
> > _Enter key functionality for selecting rows is still missing here, since its also missing in the checkbox itself, which I'll create a ticket for_
> 
> The expected key for selecting a checkbox is space, not enter. ([source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#description), also see html checkbox default behaviour)

oops! that makes sense! thanks, i'll close the ticket i made

matthewlyle (approved):

PR #496, opened 2019-12-17
Nds 1251 pick a time

marboro92:
## Description

Adds Time Picker component.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


matthewlyle:
There's a misalignment between the left padding of the input and the options themselves: 

Screen Shot 2019-12-18 at 9 15 15 AM

Also I think we should limit the minWidth to be closer to the the size of XX:XX:YY. I know this implementation matches the design file but it feels like an oversight. @nikola-nulogy thoughts?

docs/src/pages/components/time-picker.js
```diff
@@ -0,0 +1,72 @@
...
+} from "@nulogy/components";
+import Highlight from "react-highlight";
+import {
+  Layout,
+  Intro,
+  IntroText,
+  DocSection,
+  PropsTable
+} from "../../components";
+import radioProps from "../../shared/radioProps";
```
matthewlyle: this doesn't seem right. and i actually just noticed it's on the datepicker docs too. (https://nulogy.design/components/date-picker/)

components/src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,139 @@
...
+      this.setState(
+        {
+          selectedTime: date
+        },
+        () => this.handleSelectedDateChange(date)
+      );
+    }
+  };
+
+  setSelectedTimeOnInputChange = debounce(1000, value => this.setSelectedTime(getNewDateWithTime(value)));
```
matthewlyle: how short in duration can this be? the delay felt quite noticeable to me when i tried typing it. i reduced it to 200 and it felt better, but i'm not sure if that could cause an issue or if it could be even faster.

matthewlyle (commented):
it feels like something else should happen with the keyboard after typing the time. either it closing after it selects the time, or enter closes it? the [Select](https://storybook.nulogy.design/?path=/story/select--select) uses enter and it feels pretty nice so staying consistent with that would probably make sense.

docs/src/pages/components/time-picker.js
marboro92: oops totally forgot to add the props tables for both :/  will do that now

components/src/TimePicker/TimePicker.js
marboro92: I added the delay because otherwise it starts formatting while typing and it gets pretty confusing when you're backspacing, as it starts changing the input value before you're done typing.

matthewlyle: ah yeah i was just able to replicate that pretty easily with the faster version :(

matthewlyle (approved):

PR #503, opened 2020-01-06
feat(DateRange): pick a range of dates 

marboro92:
## Description

- Adds a DateRange component
- Adds a RangeContainer component for internal use that accepts the input components from the start and end of the range, and wraps the label, range "-" and error messages that is reused in the DateRange and MonthRange and should be useful for the upcoming TimeRange as well
- Cleans up the tests with helpers for mocking the currentDate and a shared custom date matcher that are used in the DateRange and MonthRange test (*Still need to skip storyshots right now since this only works for the by explicitly setting the date in the spec tests - I plan to continue that work in the NDS-1286)
- Adds highlighting for selected dates (visually indicate start date when selecting end date and vice versa)

## Changes include

- [ ] Documentation only

## Checklist

- [x] e2e tests added for component iterations

## Before Merging

### DIFF
--- a/components/src/DateRange/DateRange.js
+++ b/components/src/DateRange/DateRange.js
@@ -1,10 +1,11 @@
 import React, { useState, useEffect } from "react";
 import PropTypes from "prop-types";
-import { isBefore } from "date-fns";
+import { isBefore, addDays, differenceInDays } from "date-fns";
 import { DatePicker } from "../DatePicker";
 import { RangeContainer } from "../RangeContainer";
 import { InputFieldPropTypes, InputFieldDefaultProps } from "../Input/InputField.type";
 import { FieldLabelDefaultProps, FieldLabelProps } from "../FieldLabel/FieldLabel.type";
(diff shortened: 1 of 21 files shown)


components/src/DateRange/DateRange.js
```diff
@@ -0,0 +1,160 @@
...
+
+  const getAllDaysInRange = () => {
+    if (endDate && startDate && isBefore(startDate, endDate)) {
+      const days = Array(differenceInDays(new Date(endDate), new Date(startDate)) + 1);
+      return days.fill(0).map((_, i) => addDays(new Date(startDate), i));
+    }
+    return [];
+  };
+
+  const highlightDates = [
```
matthewlyle: should this be in the styles file?

matthewlyle (approved):

components/src/DateRange/DateRange.js
marboro92: Yes actually that would be nice, moving it

PR #509, opened 2020-01-15
feat: add option to pass locale to all date components

marboro92:
## Description

Adds a `locale` prop to all date components.
- registers all the locales available in date-fns (we may only need a subset but for now I import all of them)
- adds a knob to each of the date components' stories so that its easy to try out different locales and see all the available options that can be passed to the `locale` prop

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


matthewlyle (approved):

docs/src/shared/localeProp.js
```diff
@@ -0,0 +1,9 @@
+const localeProp = {
+  name: "locale",
+  type: "string",
+  defaultValue: "undefined",
+  description:
+    "One of the supported locale strings: 'enUS' (US English), 'de' (German), 'fr' (French), 'es' (Spanish), 'ptBR' (Brazil Portuguese), 'da' (Dutch), 'pl' (Polish)"
```
marboro92, line 6: These could probably eventually go into a docs page on localization (?) But for now I guess its most convenient along side the prop so I've put it here.

components/src/utils/datePickerLocales.js
```diff
@@ -0,0 +1,7 @@
+import { registerLocale } from "react-datepicker";
+import { enUS, de, fr, es, ptBR, da, pl } from "date-fns/locale";
+
+export const supportedDateLocales = { enUS, de, fr, es, ptBR, da, pl };
```
marboro92, line 4: There are fewer languages then we actually had in the list since date-fns doesn't make a distinction between french (fr) and spanish (es) in different countries.

docs/src/shared/localeProp.js
matthewlyle: Yeah we'll do a docs page on localization soon. For now this is good.

matthewlyle (approved):

PR #511, opened 2020-01-17
feat: add auto-complete functionality to month picker input

marboro92:
## Description

Add auto-complete functionality to the month picker input:
- When a user types in at least three letters the matching month is selected and the year is filled in with the last selected year. ex: if the original selection is “Jan 2019”, and typing in “jul” should auto fill to “Jul 2019”

- Added a disableAutoComplete prop in case this functionality doesn't work for some implementations (ex: passing in a totally different date format than the default)

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/components/src/MonthRange/MonthRange.js
+++ b/components/src/MonthRange/MonthRange.js
@@ -23,7 +23,7 @@ const MonthRange = ({
   minDate,
   maxDate,
   locale,
-  disableAutofill
+  disableAutoComplete
 }) => {
   const [startMonth, setStartMonth] = useState(defaultStartDate);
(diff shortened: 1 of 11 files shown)


components/src/MonthRange/MonthRange.js
```diff
@@ -22,7 +22,8 @@ const MonthRange = ({
   labelProps,
   minDate,
   maxDate,
-  locale
+  locale,
+  disableAutoComplete
```
matthewlyle: i think this should be cased `disableAutocomplete` to stay consistent with `autocomplete` in Select

matthewlyle (approved):

PR #518, opened 2020-01-22
feat: navigate calendars with up and down keys

marboro92:
## Description
Navigate the date picker using arrow keys

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


matthewlyle:
1) Using up and down keys works well when you've already selected a date, but if no date is selected, and the Datepicker is open, up and down do nothing. Maybe they could start by hovering over the first date of the month or today's date or something? Google Calendar gets around this because you always have a starting date for their event. 

2) How do you actually select the date with your keyboard? Space and enter both do nothing. To reference Google Calendar again: they are adding :focus styles when you're navigating with keyboard, but enter selects the date.

matthewlyle:
Navigating looks good, but now I'm wondering... how do you open the datepicker in the first place with a keyboard? (Select is with the down key if we want to keep that pattern)

matthewlyle (approved):

PR #529, opened 2020-01-29
feat!: wrap popper in reusable component and use in Popper and Dropdown Menu

marboro92:
## Description
Created Popper component to be reused wherever Popper is used. External API's for tooltip and dropdown should be left unchanged.

withMenuState HOC is now only used in NavBar.

## Changes include

- [ ] Documentation only
- [x] Refactor

## Checklist

## Before Merging


components/src/Popper/Alert.story.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
```
matthewlyle: i think this whole file being moved is a mistake?

marboro92:
Reminder to myself to mark this as a **feat!:** so we can capture the breaking change of the menuState prop being removed (in case anyone figured out they could access it.. but I don't think so)

matthewlyle (approved):

PR #557, opened 2020-02-20
feat: set up locale option on NDSProvider

marboro92:
## Description

- Set up react-i18n configuration
- Set up json files for each locale
- add a prop to NDSProvider that accepts a locale string to change the json resource that is used for string translations in components
- Changed strings to translatable strings (more of these updates to come)
- Add knob with label "NDSProvider Locale" to easily check updates when changing between locales

## Changes include

- [ ] Bugfix (non-breaking change that solves an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)
- [ ] Documentation only

## Checklist

**Please check all that apply.**

- [x] Storybook updated with examples of new functionality
- [ ] Storybook uses variable and realistic data (ex: short and long text)
- [ ] Docs updated with correct props and examples
- [ ] Updated and reviewed changes to storyshots
- [ ] e2e tests added for component interations
- [ ] jest tests added for component API that may not be captured with storyshots (change handlers, renderers etc)
- [ ] Accessibility (includes relevant tags, keyboard functionality, colour contrast)

## Before Merging

- [ ] Tested storybook deployment preview
- [ ] Tested docs deployment preview

PR #570, opened 2020-02-27
feat: add localization to nds

marboro92:
## Description

A fork continuation of https://github.com/nulogy/design-system/pull/557 (after clearing cache for that branch in Travis the build kept failing) 

- Set up react-i18next configuration
- Set up json files for each locale
- add a prop to NDSProvider that accepts a locale string to change the json resource that is used for string translations in components
- Changed strings to translatable strings (more of these updates to come)
- Add knob with label "NDSProvider Locale" to easily check updates when changing between locales

Storyshots was found to be causing a memory leak on CI after adding react-i18next. To fix this issue I changed the renderer setting for storyshots from mount to render. This removed a large bulk of the NDSProvider theme and locale setting from all the storyshots which actually makes them alot more readable.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


components/src/Alert/__snapshots__/Alert.story.storyshot
```diff
@@ -2,3880 +2,237 @@
...
 exports[`Storyshots Alert Danger 1`] = `
 <div
-  style={
-    Object {
-      "padding": "24px",
-    }
-  }
+  style="padding:24px"
 >
-  <NDSProvider
```
marboro92, line 11: Changing from mount to render for the storyshots renderer removed all the theme data and the locale object from each storyshot

.travis.yml
```diff
@@ -26,9 +27,7 @@ before_deploy:
 deploy:
   provider: script
   script: yarn lerna publish --yes
-  email: marianneb@nulogy.com
```
marboro92, line 29: Ended up cleaning these up while trying to fix the memory leak issue but they had nothing to do with it.  I just noticed that Travis had a notification that these fields were unknown or deprecated anyways.

components/src/Icon/Icon.story.js
```diff
@@ -21,7 +21,7 @@ IconCode.propTypes = {
 };
 
 storiesOf("Icon", module)
-  .add("Icon", () => (
+  .add("Icon (SkipStoryshot)", () => (
```
marboro92, line 24: After switching to the render option for storyshots somehow  the icons in the storyshot on CI were in a slightly different order than on my local and that caused failures. Opting to skip them for now.. since the storyshots are just outputting each of the svgs.

nikola-nulogy (approved):

PR #583, opened 2020-03-17
feat: add truncated text component

marboro92:
## Description

Adds a TruncatedText component which truncates text when the text is longer than maxCharacters. Optionally displays a tooltip with the full text on hover. 

Currently only supports strings without nested dom nodes within it. For example, supports "some text to be truncated" but not "some text to be` <i>truncated</i>`"

Screen Shot 2020-03-17 at 12 09 52 PM

## Changes include

- [ ] Documentation only

## Checklist

  - no additional pieces to capture outside of e2e tests

- [x] Accessibility (includes relevant tags, keyboard functionality, colour contrast)

## Before Merging

### DIFF
--- a/components/src/TruncatedText/TruncatedText.js
+++ b/components/src/TruncatedText/TruncatedText.js
@@ -1,0 +1,55 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import { Tooltip } from "../Tooltip";
+import { Text } from "../Type";
+
+const StyledTruncatedText = styled(Text)(({ hoverable }) => ({
+  width: "fit-content",
+  cursor: hoverable ? "pointer" : "default"
+}));
+
+const MaybeTooltip = ({ children, showTooltip, ...props }) => {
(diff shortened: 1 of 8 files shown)


components/src/TruncatedText/TruncatedText.js
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
+import { Tooltip } from "../Tooltip";
+import { Text } from "../Type";
+
+const StyledTruncatedText = styled(Text)(({ hoverable }) => ({
```
matthewlyle: i now realize this was missing from the acceptance criteria, but the text that gets truncated is sometimes (and often) a title, so there should be a way to pass something other than Text. i don't think it needs to be generic, just the ability for it to at least be Text, Title or SectionTitle... maybe by passing the `as` prop down to here and then documenting the possibility?

matthewlyle (approved):

PR #585, opened 2020-03-19
feat: add Toast component

marboro92:
## Description

Adds a toast component and an example of multiple toasts.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/components/src/Toast/Toast.js
+++ b/components/src/Toast/Toast.js
@@ -1,0 +1,121 @@
+import React, { useState, useEffect } from "react";
+import styled from "styled-components";
+import PropTypes from "prop-types";
+import theme from "../theme";
+import { Box } from "../Box";
+import { Alert } from "../Alert";
+
+const ANIMATION_DURATION = 2000; // in ms
+const TOAST_Y_MAX = "0px";
+const TOAST_Y_MIN = "-32px";
+const ACTIVE_Z_INDEX = 2;
+const INACTIVE_Z_INDEX = ACTIVE_Z_INDEX - 1;
+
+const SLIDE_IN_STYLES = {
+  transform: `translateY(${TOAST_Y_MIN})`,
+  transition: "transform 0.15s ease-out",
+  zIndex: ACTIVE_Z_INDEX
+};
+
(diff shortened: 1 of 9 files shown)


components/src/Toast/Toast.story.js
```diff
@@ -0,0 +1,110 @@
...
+    };
+    return <ToastWithTrigger />;
+  })
+  .add("multiple toasts example", () => {
+    const MultipleToastsExample = () => {
+      const [currentToast, setCurrentToast] = useState(undefined);
+      const triggerToast = toastName => {
+        setCurrentToast(toastName);
+      };
+      /* Alternate trigger toast fn adds a delay between toasts
```
marboro92, line 34: I'm not sure about whether this delay is needed. I tried but waiting for the original toast to fade out makes clicking the buttons feel less responsive.. lmk your thoughts you can uncomment it and comment out the other function to try it out.

components/src/Toast/Toast.js
```diff
@@ -0,0 +1,121 @@
...
+};
+
+const FADE_OUT_STYLES = {
+  transition: "opacity 1s linear"
+};
+
+const AnimatedAlert = styled(Alert)(({ visible }) => ({
+  boxShadow: theme.shadows.medium,
+  minWidth: "200px",
+  maxWidth: "600px",
```
marboro92, line 38: just using this max width to restrict the characters per line.. I'm not doing anything to enforce the number or lines/max text in the code, I opted for just adding that usage info in docs.
 It seems like that would be a confusing if the component cut off text that was inputted and would make it difficult to deal with longer translations

nikola-nulogy: This makes sense to me.

components/src/Toast/Toast.story.js
nikola-nulogy: Honestly, I can't tell the difference and I am good with either option.

nikola-nulogy (changes requested):
Title on the doc site it Tooltip

nikola-nulogy (approved):

PR #595, opened 2020-03-30
feat: pass locale to react datepicker based on setting on NDSProvider

marboro92:
## Description

Use context to pass the locale set on NDSProvider down to components relying on date-fns. Previously, a locale had to be passed to each of these components MonthPicker, MonthRange, TimePicker, TimeRange, DateRange, DatePicker.

Production Scheduling seems to be the only app using our datepicker currently, but they don't use the locale prop so removing it shouldn't cause any problems.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


matthewlyle:
Screen Shot 2020-03-31 at 8 05 38 AM
For some reason only German and English have capitals to start their month names. Is this correct or do we need to force capitalization here? Looked weird to me, but I only speak English 🤷‍♀

matthewlyle (approved):

PR #599, opened 2020-04-02
feat: create radio css component

matthewlyle:
## Description

Adds a CSS radio button.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/css/src/scss/components/_checkbox.scss
+++ b/css/src/scss/components/_checkbox.scss
@@ -7,7 +7,7 @@
   width: auto;
   user-select: none;
   align-items: center;
-  padding: $size-base-half 0;
+  padding: $size-base-x-1 0;
 }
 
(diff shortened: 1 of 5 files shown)


css/src/scss/components/_checkbox.scss
```diff
@@ -7,7 +7,7 @@
   width: auto;
   user-select: none;
   align-items: center;
-  padding: $size-base-x-1 0;
+  padding: $size-base-half 0;
```
marboro92: I used 8px here because it matches the spacing between checkboxes in the React components. In the React component has the label has 4px padding and an extra outer div that adds another 4px spacing = 8px. I think a similar thing happens for the Radio.

marboro92 (approved):

PR #613, opened 2020-04-06
feat: add iconic button component

marboro92:
## Description

Add iconic button. For the story examples, the material icon font is loaded instead of svg's to avoid needing to set up an assets folder.

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/css/src/scss/components/_iconic-button.scss
+++ b/css/src/scss/components/_iconic-button.scss
@@ -1,0 +1,36 @@
+/* -----------------------------------------
+    ## Iconic Button
+----------------------------------------- */
+
+.IconicButton {
+  display: inline-flex;
+  border: none;
+  justify-content: center;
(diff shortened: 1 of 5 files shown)


css/src/scss/components/_iconic-button.scss
```diff
@@ -0,0 +1,36 @@
...
+    box-shadow: $shadow-focus;
+  }
+  &:hover i {
+    background: $color-base-light-blue;
+  }
+  &:active i {
+    transform: scale(0.875);
+    transition: 0.2s ease-in;
+  }
+  i {
```
matthewlyle: we should probably give this a class in case people use other elements, eg svg or img, since it would still fit our naming convention. something like `IconicButton__Icon`

matthewlyle (approved):

PR #631, opened 2020-04-21
fix: remove max width from Alert to match React component

marboro92:
## Description

Just noticed bw our css alert  isn't full width like the react component. 

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/css/src/scss/components/_alert.scss
+++ b/css/src/scss/components/_alert.scss
@@ -19,7 +19,6 @@
 
 .Alert {
   @include alert($color-base-blue, $color-base-light-blue);
-  display: flex;
 }
 
 .Alert__title {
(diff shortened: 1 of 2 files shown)


css/src/scss/components/_alert.scss
```diff
@@ -20,7 +20,6 @@
 .Alert {
   @include alert($color-base-blue, $color-base-light-blue);
   display: flex;
```
matthewlyle: We were talking about removing the Flex too. Do you think we should do that now?

marboro92: We were using that flex to position the icon within the Alert but we could update the examples with icons to use the utility class .nds-flex?
Though we could also try updating the helpers in Pack Manager to include those alert icons if we want to match the react component that closely.

matthewlyle: Yeah I think we should update the icon examples to use flex instead of having it always be flex. Otherwise I think it's kind of a weird experience to see the content automatically columned out. It's easier in React since we control the markup =/

PR #632, opened 2020-04-22
fix: add module to package and remove file restriction to support auto-imports

marboro92:
An attempt to get auto-imports to work in RubyMine:

Adding a module file to package.json seems to fix the issue where the imports are coming from internal files in RubyMine. All components get imported from the package "@nulogy/components" rather than "@nulogy/components/src/{componentName}" when tested locally or by publishing a new version under the next tag.

info about module setting: https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#typical-usage

## Description

**Include what change you're making, why, and link to any relevant JIRA issues**

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging


components/package.json
```diff
@@ -41,10 +41,8 @@
     "node": "*"
   },
   "main": "dist/main.js",
+  "module": "src/index.js",
```
marboro92: I believe we could replace with with a generated es module in /dist and maintain the same auto-complete functionality while also being able to package only the dist folder (without src). Other libraries I've looked at manage to do that using Rollup, BUT.. Webpack doesn't support es modules yet https://github.com/webpack/webpack/issues/2933 but they might soon

marboro92:
TO DO: 
-  [x] clean up or find easier work around for having to specify all named exports in Rollup? 
- [ ] remove manual versioning and add a migration guide

TO TEST: 
- [x] auto-imports work in RubyMine with new main.module set up and only packaging dist folder
- [x] prop-type warnings still work in dev mode 
- [x] consuming app can consume components without run-time errors
        - currently styled-components has to be loaded as a peerDependency for the rollup build to work but wasn't loaded that way with webpack..
- [x] nulogy.design site works as expected without a seperate target: "node" build
- [x] investigate whether or not to load styled-components as a peerDependency 

TESTED BY CONSUMING APPS:
- [x] QCloud
- [x] Go

marboro92:
moved to  branch for v3 beta version

PR #639, opened 2020-05-04
feat: add PM style table

marboro92:
#  Description

Add a PM style table through the boolean prop packManagerStyle. I've used the css prop heavily to make these updates to avoid updating the code in too many places and make these style overrides easy to remove when the time comes. 

## Current PM:
PM-current-table

Screen Shot 2020-05-04 at 2 11 15 PM

## NDS in PM Style: 
Tables:
Screen Shot 2020-05-04 at 5 14 19 PM

Pagination:
Screen Shot 2020-05-04 at 5 15 16 PM

## Changes include

- [ ] Documentation only

## Checklist

## Before Merging

### DIFF
--- a/components/src/Icon/LoadingIcon.js
+++ b/components/src/Icon/LoadingIcon.js
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { generateId } from "../utils";
+import generateId from "../utils/generateId";
 
 const LoadingIcon = ({ color, size, title, ...props }) => {
   const id = generateId();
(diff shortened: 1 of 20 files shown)


marboro92:
~I wonder if rather than adding a prop option for nds we use the the styled-components css prop for our components as needed and provide the css object to override NDS styles to something more like packmanager (see Pagination.story)?~

~This would allow us to use variables defined in pack-manager within the override styles and perform gradual upgrades across all implementations of tables by updating variables just like we plan to for fonts and background colors.~

We decided to provide as simple of a solution as possible thats intended to be temporary and should be replaced to follow the normal design guidelines.

TODO: 

- [x] left and right icons in pagination
- [x] add PM version of pagination component
- [x] add override for pagination built-in table
- [x] use more realistic data resembling whats in packmanager for the examples

matthewlyle (approved):

PR #707, opened 2020-07-02
feat: support using typescript for components

marboro92:
## Description
To try this out you can..
- Run `yarn build` and look at the resulting `dist` folder and see type declarations under `src`
- Convert components in .ts files by specifying and using a type

## Changes include

- [ ] Documentation only

## Checklist

- [ ] e2e tests added for component iterations

## Before Merging


components/.storybook/main.js
```diff
@@ -7,5 +7,21 @@ module.exports = {
...
     "@storybook/addon-actions",
     "storybook-addon-performance"
-  ]
+  ],
+  webpackFinal: async config => {
+    config.module.rules.push({
+      test: /\.(ts|tsx)$/,
+      use: [
+        {
+          loader: require.resolve("ts-loader")
```
ecbrodie: I normally recommend using Babel's typescript preset over `ts-loader`. That way, you can just stick to using `babel-loader`, just like if you were using vanilla JS. https://babeljs.io/docs/en/babel-preset-typescript

But it's just a suggestion. If you prefer `ts-loader`, then stick to that.

components/babel.config.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
         useBuiltIns: "entry"
       }
     ],
-    "@babel/preset-react"
+    "@babel/preset-react",
+    "@babel/preset-typescript"
```
ecbrodie: Hmm...looks like you already use Babel's typescript preset, so you really don't need `ts-loader`.

PR #748, opened 2020-08-21
feat: convert our components to typescript

marboro92:
## Description

Updates our components to use .tsx. Component api's should all support autocomplete now and external API's are all defined now with simple types. 

Notes:
- tried moving from typescript2 to plugin-typescript which caught more typescript issues but I ran into issues trying to configure builds in two formats, so leaving that out for now https://github.com/rollup/plugins/issues/247

## Changes include

- [ ] Documentation only

## Checklist

- [ ] e2e tests added for component iterations

## Before Merging


package.json
```diff
@@ -50,8 +49,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn run precommit",
```
marboro92, line 53: has opposite settings to our lint configurations that are checked in pre-push which caused me to be unable to push

matthewlyle (approved):

package.json
matthewlyle: should we delete the task and dependencies too then?

PR #773, opened 2020-09-23
chore: Add Chromatic workflow

matthewlyle:
## Description

Adds a workflow to deploy Chromatic...

- [ ] Test that a change to something in the component src will trigger Chromatic
- [x] Test that a change to something not in the component src will not trigger Chromatic


marboro92 (approved):
testing out chromatic build

PR #806, opened 2020-11-06
feat: add a wrench to `Icon` component

matthewlyle:


marboro92:
We're ahead of this version now

PR #826, opened 2020-12-03
fix(AsyncSelect)!: AsyncSelect can now be used as a controlled compo…

marboro92:
## Description

AsyncSelect can now be used as a controlled component by setting the `value` prop. `onChange` now returns the complete option object that is selected rather than just the string value.

Here's an example of migrating an existing onChange handler: 

```
onChange = (newValue: string) => {
     setValue(value)
}

would need to change to 

onChange = (newValue: <{value: string,  label}>) => {
    setValue(newValue.value);
}
```
## Changes include

## Feature checklist


marboro92:
will go to beta branch

marboro92:
To be included in v5

PR #834, opened 2020-12-22
fix: allow variable width of daterange component

marboro92:
## Description
- change styles on range container so that it will wrap the fields when there isn't enough space
- add ability to set inputWidth on the datepicker inputs in DateRange

## Changes include

## Feature checklist


marboro92:
TODO: wrap date and time fields together instead of one by one

PR #863, opened 2021-02-26
feat(IconicButton): Adds ability to set a color

danielsilva:
## Description

**This is to add ability to configure the color of IconicButtons.**

**For example:**
image

## Changes include

- [X] feature: a non-breaking change that adds functionality

## Feature checklist

### DIFF
--- a/src/Button/IconicButton.story.js
+++ b/src/Button/IconicButton.story.js
@@ -1,7 +1,6 @@
 import React from "react";
 import { IconicButton } from "../index";
 import { Flex } from "../Flex";
-import theme from "../theme";
 
 export default {
   title: "Components/IconicButton",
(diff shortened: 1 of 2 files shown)


src/Button/IconicButton.story.js
```diff
@@ -65,6 +66,16 @@ WithACustomSize.story = {
   name: "with a custom size",
 };
 
+export const WithACustomColor = () => (
+  <IconicButton color={theme.colors.red} icon="close" labelHidden>
```
matthewlyle: You don't need to pull the theme in here. Since storybook is wrapped in the `NDSProvider`, theme values can be used directly, e.g `<IconicButton color="red" />`.

PR #919, opened 2021-07-16
feat: Add option to turn off resizing in a Textarea

matthewlyle:
## Description

Textareas are resizeable by default. With CSS, you can add `{ resize: none }` to prevent that. Since it’s a CSS style, and not an HTML tag, our `Textarea` styled-component doesn’t allow you to do that. This PR adds that functionality.

## Changes include

## Feature checklist

### DIFF
--- a/src/Textarea/Textarea.tsx
+++ b/src/Textarea/Textarea.tsx
@@ -1,5 +1,5 @@
 import React, { forwardRef } from "react";
-import styled from "styled-components";
+import styled, { CSSObject } from "styled-components";
 import propTypes from "@styled-system/prop-types";
 import { transparentize } from "polished";
 import { SpaceProps } from "styled-system";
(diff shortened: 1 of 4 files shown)


src/Textarea/Textarea.tsx
```diff
@@ -44,10 +44,11 @@ type StyledTextareaProps = React.ComponentPropsWithRef<"textarea"> &
     errorList?: string[];
     error?: boolean;
     rows?: number;
+    isResizeable?: boolean;
   };
 const StyledTextarea: React.SFC<StyledTextareaProps> = styled.textarea(
   space,
-  ({ theme }) => ({
+  ({ theme, isResizeable }: any): CSSObject => ({
```
haideralsh, line 51: I am wondering if we could've used `StyledTextareaProps` instead of `any` here to make use of TS's type checking if someone accidentally puts a prop there that is not part of `StyledTextareaProps`?

matthewlyle: Oops sorry i missed this comment. Yes good call. We can push another release for that.

PR #921, opened 2021-07-21
fix: Allow clearing all items in the AsyncSelect component

haideralsh:
## Description

The AsyncSelect component hides the dropdown button and the clear button using an attribute selector CSS. This PR removes it in favour of a the react-select [config](https://react-select.com/props#prop-types) that only hides the dropdown button without affecting the clear button.

## Changes include

## Feature checklist

### DIFF
--- a/cypress/integration/components/AsyncSelect.spec.js
+++ b/cypress/integration/components/AsyncSelect.spec.js
@@ -29,7 +29,7 @@ describe("AsyncSelect", () => {
       getMultiselect().click();
 
       cy.focused().type("cana");
-      cy.wait(200);
+      assertDropDownIsOpen().contains("Canada");
       cy.focused().type("{enter}");
 
       getClearButton().click();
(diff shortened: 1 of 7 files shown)


cypress/integration/components/AsyncSelect.spec.js
```diff
@@ -0,0 +1,48 @@
...
+      assertDropDownIsOpen().contains("Canada");
+      cy.focused().type("{enter}");
+
+      getClearButton().click();
+
+      getMultiselect().contains("Please select a countries");
+    });
+
+    it("does not show the dropdown arrow", () => {
+      getDropdownMenu().should("not.exist");
```
matthewlyle: Is `getDropDownMenu` what we're trying to test here or should we have a data-test id on the arrow itself to just check if it's showing?

haideralsh: You're right, we're not asserting the right thing.

haideralsh: I added a new test that asserts that the dropdown arrow is not showing using its data-testid.

matthewlyle (approved):

PR #923, opened 2021-07-23
fix: Remove @hypnosphi/create-react-context dependency 

matthewlyle:
## Description

In a recent release, our Popper version updated from 1.3.7 to 1.3.11:

before
after

In this change, the dependency on `create-react-context` changed to `hypnosphi/create-react-context`. This caused an issue in Eco for some reason, where they'd get a missing dependency error. 

Popper v1 is already a legacy dependency we'd love to get rid of, so I don't want to add any code to support changes they make in that library. Instead, I've locked our v1 Popper version to 1.3.7 so it doesn't make us think about it again. Once I fixed that, I noticed that react-datepicker suffered from the same issue (as it also used Popper v1), so I upgraded the dependency to the latest version, which uses Popper v2. We had to pass two breaking versions and they don't publish a changelog (???) but all of our tests are still passing so we _should_ be good...

## Changes include

## Feature checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
     "framer-motion": "^3.1.3",
     "i18next": "^19.3.1",
     "polished": "3.4.4",
-    "react-datepicker": "^2.10.1",
+    "react-datepicker": "^4.1.0",
     "react-fast-compare": "^3.2.0",
     "react-i18next": "^11.3.3",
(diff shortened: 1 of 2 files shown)


haideralsh:
LGTM! 👍

haideralsh (approved):

PR #926, opened 2021-08-04
fix: Add spacing to the Toast close button and align it with text 

haideralsh:
## Description

Fixes the issue with the close button in the `<Toast />` component being vertically misaligned with the text and sitting super close to the text.  

**Describe the change you're making, the motiviations behind it, and any thing else a reviewer may need to know to approve this PR.**

## Changes include

## Feature checklist

### DIFF
--- a/src/Alert/Alert.tsx
+++ b/src/Alert/Alert.tsx
@@ -2,23 +2,23 @@ import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
 import styled from "styled-components";
 import { Box } from "../Box";
-import {BoxProps} from "../Box/Box";
+import { BoxProps } from "../Box/Box";
 import { Icon } from "../Icon";
 import { Link } from "../Link";
 import { Flex } from "../Flex";
 import { Text } from "../Type";
 
 export type AlertProps = BoxProps & {
-    children?: React.ReactNode;
-    className?: string;
-    isCloseable?: boolean;
-    closeAriaLabel?: string;
-    title?: string;
-    type?: "danger" | "informative" | "success" | "warning" | undefined;
-    onClose?: any;
-    controlled?: boolean;
-    style?: React.CSSProperties;
-  };
+  children?: React.ReactNode;
+  className?: string;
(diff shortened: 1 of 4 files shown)


src/Alert/Alert.tsx
```diff
@@ -2,23 +2,28 @@ import React, { useState } from "react";
...
+  isCloseable?: boolean;
+  closeAriaLabel?: string;
+  title?: string;
+  type?: "danger" | "informative" | "success" | "warning" | undefined;
+  onClose?: any;
+  controlled?: boolean;
+  style?: React.CSSProperties;
+};
+
+const CollapsedLink = styled(Link)<LinkProps>(() => ({
```
matthewlyle: Did you add this component because the Link didn't accept a `lineHeight` prop? We should probably add the prop to the Link instead of just adding CSS so it can be leveraged in other cases as well.

haideralsh: Good point. I see `typography` is already available on the `Link` component, but the `TypographyProps` is not, that's why I was getting the error when adding it directly on the the `Link`. 

I added `TypographyProps` to the `LinkProps` and removed the styled component.

src/Toast/Toast.story.tsx
```diff
@@ -178,6 +178,36 @@ WithCloseButton.story = {
   name: "with close button",
 };
 
+export const CloseableWithLongMessage = () => {
```
matthewlyle: Do you see any value in having both a `Closeable` story and a `CloseableWithLongMessage` or should we just make the example in the original Closeable always have a long message?

haideralsh: I guess not in this case, but the `CloseableWithLongMessage` could _actually_ have a long message with 150 character limit outlined in the [guideline](https://nulogy.design/components/toast) in case a future change breaks multiline Toasts. What do you think?

matthewlyle: Good idea.

matthewlyle (approved):

PR #966, opened 2021-11-15
feat: page header improvements

haideralsh:
## Description
Adds the updated header component alongside the summary, and introduces some DX and UI improvements.

## Changes include

## Feature checklist

### DIFF
--- a/src/Layout/Header.story.tsx
+++ b/src/Layout/Header.story.tsx
@@ -1,0 +1,176 @@
+import React from "react";
+import { Box, Breadcrumbs, Button, DropdownButton, DropdownLink, Flex, Link, PrimaryButton, StatusIndicator } from "..";
+import Summary from "../Summary/Summary";
+import SummaryItem from "../Summary/SummaryItem";
+import SummaryDivider from "../Summary/SummaryDivider";
+import DropdownMenu from "../DropdownMenu/DropdownMenu";
+import Header from "./Header";
+
+export const DefaultBackground = () => (
+  <Header
+    renderBreadcrumbs={() => (
+      <Breadcrumbs>
+        <Link href="/">Home</Link>
+        <Link href="/">Materials</Link>
+      </Breadcrumbs>
+    )}
+    title="Materials Overview"
+  />
+);
+
+export const WithAnUndecoratedBackground = () => (
+  <Header
+    undecorated
+    renderBreadcrumbs={() => (
+      <Breadcrumbs>
+        <Link href="/">Home</Link>
+        <Link href="/">Materials</Link>
+      </Breadcrumbs>
+    )}
+    title="Materials Overview"
+  />
+);
+
+export const WithActions = () => (
+  <Header
+    undecorated
+    renderBreadcrumbs={() => (
+      <Breadcrumbs>
+        <Link href="/">Home</Link>
+        <Link href="/">Materials</Link>
+      </Breadcrumbs>
+    )}
+    title="Materials Overview"
+    renderActions={() => (
+      <Flex gap="x1">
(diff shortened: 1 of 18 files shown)


src/Layout/Header.tsx
```diff
@@ -0,0 +1,116 @@
...
+              )}
+
+              {children && <Box flexGrow={1}>{children}</Box>}
+            </Flex>
+          </Flex>
+          <Flex display="flex" alignItems="flex-start" gap="x2">
+            {summary &&
+              !m &&
+              !sm &&
+              React.cloneElement(summary, {
```
petror-nulogy: If you do not have special reasons to use `React.cloneElement`, I would select [render prop](https://beta.reactjs.org/apis/react/cloneElement#passing-data-with-a-render-prop) instead.

src/Layout/Header.tsx
```diff
@@ -0,0 +1,116 @@
...
+        height="45%"
+        background="linear-gradient(173.35deg, #E0E3E6 5.22%, rgba(205, 219, 237, 0) 104.74%)"
+        opacity="0.5"
+      />
+      <RightAngleTriangle height="20%" bg="darkblue" opacity="0.05" />
+    </Box>
+  );
+};
+
+const Header: React.FC<HeaderProps> = ({
```
petror-nulogy, line 25: With ECO we propose to move all additional components down of the file.
At the top of file leave only main component and props of main component.
What do you think about it? :)

src/Layout/Header.tsx
petror-nulogy: But in case if Header is available as NDS component, I agree that for end user provide component instead render prop will be more convenient.

src/Summary/SummaryItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <SummaryItemWrapper collapseBreakpoint={collapseBreakpoint}>
+      {matches ? <Text fontWeight="medium">{value}</Text> : <Heading3 mb="0">{value}</Heading3>}
+      {status}
+    </SummaryItemWrapper>
+  );
+};
+
+SummaryItem.displayName = "SummaryItem";
```
petror-nulogy: I think we can avoid `displayName` if it's the same as variable name.

src/Summary/Summary.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <SummaryContextProvider collapseBreakpoint={collapseBreakpointPx}>
+      <SummaryWrapper collapseBreakpoint={collapseBreakpointPx}>{children}</SummaryWrapper>
+    </SummaryContextProvider>
+  );
+};
+
+Summary.displayName = "Summary";
+
+Summary.defaultProps = {
+  collapseBreakpoint: DEFAULT_COLLAPSE_BREAKPOINT,
```
petror-nulogy: Does it make sense to move `defaultProps` directly to component definition? Example:
```
const Summary: React.FC<SummaryProps> = ({ collapseBreakpoint = DEFAULT_COLLAPSE_BREAKPOINT, children }) => {
```

src/Layout/Header.story.tsx
```diff
@@ -0,0 +1,112 @@
...
+    breadcrumbs={
+      <Breadcrumbs>
+        <Link href="/">Home</Link>
+        <Link href="/">Materials</Link>
+      </Breadcrumbs>
+    }
+    title="Materials Overview"
+    subtitle="Materials Overview"
+    summary={
+      <Summary>
```
petror-nulogy: Also, using `React.cloneElement` we will get hidden props like `collapseBreakpoint`. Comparing to render prop we can define a type of a function.

PR #980, opened 2022-01-06
feat: Add sorting table component

jerridan:
## Description

This adds a new SortingTable component.
We found the Table with sorting example to be broken while implementing it in Production Scheduling. So we fixed that, and decided to extract it to a new component that others could use.

## Changes include

## Feature checklist

### DIFF
--- a/src/SortingTable/SortingTable.tsx
+++ b/src/SortingTable/SortingTable.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
-import { Table, TableProps } from "../Table";
+import { Table } from "../Table";
+import type { TableProps } from "../Table";
 import type { RowType, ColumnType } from "../Table/Table.types";
 
 type SortingTableProps = TableProps & {
@@ -38,12 +39,12 @@ const sortRows = (
   return sortState.ascending ? sortedRows : sortedRows.reverse();
 };
 
-const SortingTable = ({
+const SortingTable: React.FC<SortingTableProps> = ({
   columns: incomingColumns,
   rows: incomingRows,
   initialSortColumn,
   ...props
-}: SortingTableProps) => {
+}) => {
   const [sortState, setSortState] = useState<SortState>({
     ascending: true,
     sortColumn: initialSortColumn,
--- a/src/SortingTable/SortingTable.tsx
+++ b/src/SortingTable/SortingTable.tsx
@@ -21,7 +21,7 @@ const applySort = (
   sortColumn: string,
   columns: SortableColumnType[]
 ) =>
-  rows.sort((a, b) => {
+  [...rows].sort((a, b) => {
     const column = columns.find((col) => col.dataKey === sortColumn);
     const { numeric } = column;
 
@@ -50,7 +50,7 @@ const SortingTable = ({
   });
 
   const [rows, setRows] = useState(() =>
-    sortRows([...incomingRows], incomingColumns, sortState)
+    sortRows(incomingRows, incomingColumns, sortState)
   );
 
   const onSortChange = (dataKey) => {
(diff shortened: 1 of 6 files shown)


src/Table/TableWithSorting.story.tsx
```diff
@@ -1,92 +0,0 @@
-/* eslint-disable react/prop-types */
```
haideralsh, line 1: Do you think it's worth keeping this example, just renaming it to "With Custom Sorting", to show how one would implement sorting that is not the standard in `<SortingTable />` if they needed to?

src/SortingTable/SortingTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return sortState.ascending ? sortedRows : sortedRows.reverse();
+};
+
+const SortingTable = ({
+  columns: incomingColumns,
+  rows: incomingRows,
+  initialSortColumn,
+  ...props
+}: SortingTableProps) => {
```
haideralsh: Do you mind changing this to use `React.FC<T>`

```
const SortingTable: React.FC<SortingTableProps> = ({
  columns: incomingColumns,
  rows: incomingRows,
  initialSortColumn,
  ...props
}) => {
```

src/Table/TableWithSorting.story.tsx
jerridan: I was actually wondering the same this morning. Originally we were just going to fix the example, but went the route of making the `SortingTable` component.
But I agree, it could still be useful to someone. I'll add it back in.

src/SortingTable/SortingTable.tsx
jerridan: This must be a typescript thing?

src/SortingTable/SortingTable.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState } from "react";
+import { Table, TableProps } from "../Table";
```
haideralsh: Can we split the named and type import as so: 

```suggestion
import { Table } from "../Table";
import type { TableProps } from "../Table";
```

jerridan: Looks like another typescript thing - sure I will change that.

src/SortingTable/SortingTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const applySort = (rows, sortColumn, columns) =>
+  rows.sort((a, b) => {
+    const column = columns.find((col) => col.dataKey === sortColumn);
+    const { numeric } = column;
+
+    return numericAlphabeticalSort(a[sortColumn], b[sortColumn], numeric);
+  });
+
+const sortRows = (rows, columns, sortState) => {
```
haideralsh: Are the argument and return types for these functions easy to define? Locally my TypeScript compiler is unable to infer any of them except for the `numericAlphabeticalSort` returning a number.

jerridan: I'm not familiar with typescript, but I have done prop types in that past. Going off of that, the arguments might be difficult since they aren't always the same.
Did you want to pair on it?

src/Table/TableWithSorting.story.tsx
jerridan: Done

src/SortingTable/SortingTable.tsx
jerridan: Done

src/SortingTable/SortingTable.tsx
jerridan: Done

haideralsh (approved):

PR #997, opened 2022-03-17
fix: Prevent validation errors from time when daterange does not show time

allank-nulogy:
## Description

This change fixes a bug in the `DateRange` component which is configured such that:
- `defaultStartDate` and `defaultEndDate` are provided which contain millisecond components
- `showTimes` is falsy
- An end date is selected that is the same as the start date

The expected result was that there would not be a validation error
The actual result was that there was a validation error claiming that `end date is before start date`

The fix was to fix the validation logic to ignore times on the same day when `showTimes` is falsy. 

## Changes include

- [X] fix: a non-breaking change that solves an issue

## Feature checklist

### DIFF
--- a/src/DateRange/DateRange.spec.tsx
+++ b/src/DateRange/DateRange.spec.tsx
@@ -108,7 +108,34 @@ describe("DateRange", () => {
       );
       expect(latestCall.error).toEqual("end date is before start date");
     });
-    it("returns the selected range with an error if the range is invalid", () => {
+    it("does not return an error if the end date is changed to the start date when times are not shown", () => {
+      const defaultStartDate = new Date();
+      const defaultEndDate = new Date(
+        defaultStartDate.getFullYear(),
+        defaultStartDate.getMonth(),
+        defaultStartDate.getDate() + 6
+      );
+      const onRangeChange = jest.fn();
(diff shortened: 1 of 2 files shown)


src/DateRange/DateRange.spec.tsx
```diff
@@ -108,7 +108,34 @@ describe("DateRange", () => {
...
       expect(latestCall.error).toEqual("end date is before start date");
     });
-    it("returns the selected range with an error if the range is invalid", () => {
+    it("does not return an error if the same date is selected without showing times", () => {
+      const startingDate = new Date();
+      const endingDate = new Date(
+        startingDate.getFullYear(),
+        startingDate.getMonth(),
+        startingDate.getDate() + 6
+      );
```
haideralsh, line 117: Shouldn't the ending date be the same as the starting date to match the assertion `if the same date is selected`?

allank-nulogy: To better match our motivating use case, we wanted to start with a different date, then select the same end date. That said, I think we can change our variable names to better reflect this.

allank-nulogy: and possibly change the test name to emphasize the fact that select = click instead of setup

haideralsh (approved):

PR #1005, opened 2022-05-10
fix: Ensure tab focus takes into account defaultSelectedIndex

blehoux17:
## Description

This change fixes a bug in the `Tab` component which always set the initial focus to the first tab even when a `defaultSelectedIndex` was set.

This change also fixes a similar issue in the `Switcher` component with a `selected` value.

## Changes include

## Feature checklist

### DIFF
--- a/src/Switcher/Switcher.tsx
+++ b/src/Switcher/Switcher.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { ReactElement } from "react";
 import PropTypes from "prop-types";
 import { Box } from "../Box";
 import FocusManager from "../utils/ts/FocusManager";
@@ -19,6 +19,12 @@ const Switcher: React.FC<SwitcherProps> = ({ selected, onChange, ...rest }) => {
 
     return value === selected;
   };
+
+  const getSelectedIndex = () => {
+    return React.Children.toArray(rest.children).findIndex(
+      (child) => (child as ReactElement)?.props?.value === selected
+    )
+  }
 
   const options = (focusedIndex, setFocusedIndex, handleArrowNavigation) => {
     return React.Children.map(rest.children, (child, index) => {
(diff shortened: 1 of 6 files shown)


haideralsh (approved):
LGTM. Thank you for the PR @blehoux17.

src/Switcher/Switcher.tsx
```diff
@@ -20,6 +20,12 @@ const Switcher: React.FC<SwitcherProps> = ({ selected, onChange, ...rest }) => {
     return value === selected;
   };
 
+  const getSelectedIndex = () => {
+    const options = rest.children.map((child) => child.props.value);
+
+    return options.indexOf(selected);
```
haideralsh: Should we use `findIndex` instead of `map` so that we don't loop through all elements and instead break as soon as the first child with the condition is found?

src/Tabs/Tabs.tsx
```diff
@@ -109,7 +109,7 @@ class Tabs extends React.Component<TabsProps, TabsState> {
     const spaceProps = getSubset(this.props, propTypes.space);
     return (
       <Box position="relative">
-        <FocusManager refs={this.tabRefs}>
+        <FocusManager refs={this.tabRefs} selectedIndex={this.getSelectedIndex()}>
```
haideralsh: At a second look I think this prop name looks strange. Should we rather call this `defaultFocusedIndex` instead? I just think that the focus manager shouldn't care about the selected value's index.

src/Switcher/Switcher.tsx
haideralsh: Something similar to this:

```suggestion
    return React.Children.toArray(children).findIndex(
      (child) => (child as ReactElement)?.props?.value === selected
    )
```

blehoux17: This looks great to me, I didn't know about `React.Children.toArray`

src/Tabs/Tabs.tsx
blehoux17: I agree, this is a much better name!

haideralsh (approved):

PR #1019, opened 2022-06-07
Stylistic updates

sskirby:
## Description

Executing https://nulogy-go.atlassian.net/browse/GO-4538

Implement the visual design changes from here: https://www.figma.com/file/eTgLkFDHd8GLNFmD9tBp72/Multi-Echelon?node-id=658%3A30413

Summary of changes:

1. The Tooltip has been updated to use the medium shadow.
2. For the DropdownMenu and BrandedNavBar menus and submenus (but not top-level menu triggers!):
2.1. Update shadow from small to medium.
2.2. Update background colour to white.
2.3. Update hover colour to darkBlue.
2.4. Update colour to darkGrey.
2.5. Update hover background colour to lightBlue.
3. The Sidebar shadow has been removed.

Potentially breaking changes:

1. The BrandedNavBar no longer attempts to style components rendered via the `render` property. If you want NDS styles, use an NDS component like the DropdownLink, DropdownButton, or DropdownText.

## Changes include

## Feature checklist

### DIFF
--- a/src/BrandedNavBar/MobileMenu.tsx
+++ b/src/BrandedNavBar/MobileMenu.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import PropTypes from "prop-types";
 import styled from "styled-components";
 import { display } from "styled-system";
 import { themeGet } from "@styled-system/theme-get";
@@ -7,6 +6,7 @@ import { Text, Heading3 } from "../Type";
 import { Flex } from "../Flex";
 import { BrandingText } from "../Branding";
 import { DefaultNDSThemeType } from "../theme.type";
+import { DropdownText } from "../DropdownMenu";
 import NulogyLogo from "./NulogyLogo";
 
 const borderStyle = "1px solid #e4e7eb";
@@ -17,7 +17,21 @@ const BrandingWrap = styled.div(({ theme }) => ({
   marginBottom: theme.space.x1,
 }));
 
+// eslint-disable-next-line no-mixed-operators
 const getPaddingLeft = (layer) => `${24 * layer + 24}px`;
+
+const TopLevelText = styled(Text)(({ theme }) => ({
+  color: theme.colors.blackBlue,
+  display: "block",
+  textDecoration: "none",
+  border: "none",
+  backgroundColor: "transparent",
+  fontSize: theme.fontSizes.large,
+  fontWeight: theme.fontWeights.medium,
+  lineHeight: theme.lineHeights.heading3,
+  padding: `${theme.space.x1} ${theme.space.x3}`,
+  paddingLeft: getPaddingLeft(0),
+}));
 
 const getSharedStyles = ({ color, layer, theme }) => ({
   display: "block",
@@ -110,14 +124,11 @@ const MenuLink = styled.a(
   })
 );
 
-type MenuTextProps = {
-  textColor?: string;
-  layer?: number;
-  theme?: DefaultNDSThemeType;
-};
-
-const MenuText = styled.li(({ textColor, layer, theme }: MenuTextProps) => ({
-  ...getSharedStyles({ color: textColor, layer, theme }),
+const ApplyIndent = styled.li(({ layer, theme }: MenuLinkProps) => ({
+  marginBottom: theme.space.x1,
+  [`> ${DropdownText}`]: {
+    paddingLeft: `${24 * layer + 24}px`,
+  },
 }));
 
 const SubMenuItemsList = styled.ul({
@@ -168,15 +179,14 @@ const renderSubMenu = (menuItem, linkOnClick, themeColorObject, layer) => (
   </li>
 );
 
-const renderText = (menuItem, linkOnClick, themeColorObject, layer) => (
-  <MenuText
-    key={menuItem.key ?? menuItem.name}
-    layer={layer}
-    {...themeColorObject}
-  >
-    {menuItem.name}
-  </MenuText>
-);
+const renderText = (menuItem, linkOnClick, themeColorObject, layer) => {
+  const MenuText = layer === 0 ? TopLevelText : DropdownText;
+  return (
+    <ApplyIndent layer={layer} key={menuItem.key ?? menuItem.name}>
+      <MenuText>{menuItem.name}</MenuText>
+    </ApplyIndent>
+  );
+};
 
 const getRenderFunction = (menuItem) => {
   if (menuItem.items) {
--- a/src/BrandedNavBar/MobileMenu.tsx
+++ b/src/BrandedNavBar/MobileMenu.tsx
@@ -29,6 +29,7 @@ const getSharedStyles = (theme) => ({
   fontSize: theme.fontSizes.large,
   fontWeight: theme.fontWeights.medium,
   lineHeight: theme.lineHeights.heading3,
+  marginBottom: theme.space.x1,
   padding: `${theme.space.x1} ${theme.space.x3}`,
   paddingLeft: getPaddingLeft(0),
 });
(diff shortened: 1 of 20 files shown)


src/BrandedNavBar/MobileMenu.tsx
```diff
@@ -17,107 +18,63 @@ const BrandingWrap = styled.div(({ theme }) => ({
...
-  textColor?: string;
+type ApplyIndentProps = {
   layer?: number;
   theme?: DefaultNDSThemeType;
+  key?: string;
 };
 
-const MenuText = styled.li(({ textColor, layer, theme }: MenuTextProps) => ({
-  ...getSharedStyles({ color: textColor, layer, theme }),
+const ApplyIndent = styled.li(({ layer, theme }: ApplyIndentProps) => ({
```
sskirby: @sskirby I don't like this name. It should make it obvious that it changes the child elements and it also does more than that!

src/BrandedNavBar/MobileMenu.tsx
```diff
@@ -17,107 +18,63 @@ const BrandingWrap = styled.div(({ theme }) => ({
...
+  display: "block",
+  textDecoration: "none",
+  border: "none",
+  backgroundColor: "transparent",
+  fontSize: theme.fontSizes.large,
+  fontWeight: theme.fontWeights.medium,
+  lineHeight: theme.lineHeights.heading3,
+  padding: `${theme.space.x1} ${theme.space.x3}`,
+  paddingLeft: getPaddingLeft(0),
+});
```
sskirby: @sskirby move this up to where it used to be.

src/BrandedNavBar/SubMenuTrigger.tsx
```diff
@@ -1,65 +1,33 @@
...
-import theme from "../theme";
+import styled, { StyledComponentPropsWithRef } from "styled-components";
 import { Icon } from "../Icon";
+import { DropdownButton } from "../DropdownMenu/index";
 import NavBarDropdownMenu from "./NavBarDropdownMenu";
 import renderSubMenuItems from "./renderSubMenuItems";
 
 type SubMenuTriggerProps = React.ComponentPropsWithRef<"button"> & {
   name?: string;
-  isOpen?: boolean;
```
sskirby: @sskirby why remove the `isOpen`?

src/DropdownMenu/DropdownMenu.story.tsx
```diff
@@ -75,18 +76,24 @@ WithButtonClosingMenu.story = {
...
     <DropdownItem>
       <a href="/never_been">Custom Link Component</a>
     </DropdownItem>
   </DropdownMenu>
 );
 
-WithCustomItem.story = {
+WithCustomLink.story = {
   name: "with custom item",
 };
```
sskirby: @sskirby delete this

src/BrandedNavBar/SubMenuTrigger.tsx
```diff
@@ -15,36 +15,13 @@ type SubMenuTriggerProps = React.ComponentPropsWithRef<"button"> & {
...
-      color: theme.colors.darkBlue,
-      backgroundColor: theme.colors.lightGrey,
-      borderLeft: `${theme.space.half}  solid ${theme.colors.blue}`,
-    },
-    backgroundColor: isOpen ? theme.colors.lightGrey : "transparent",
-    textDecoration: "none",
-    textAlign: "left",
-    cursor: "pointer",
+    backgroundColor: isOpen ? theme.colors.lightBlue : "transparent",
+    color: isOpen ? theme.colors.darkBlue : theme.colors.darkGrey,
```
sskirby, line 22: @haideralsh is there a better way of pulling these colour values from the `DropdownButton`? Or getting the defaults somehow?

haideralsh: I don't think we can tell the value of the props of the `DropdownButton` in this context. I think the only thing we can do is extract the values to some shared config and use it instead of the theme, but it might be overkill if this this is the only instance where we need these values.

src/BrandedNavBar/SubMenuTrigger.tsx
```diff
@@ -57,9 +34,8 @@ const SubMenuTriggerButton = React.forwardRef<
   <StyledButton isOpen={isOpen} ref={ref} {...props}>
     {name}
     <Icon
-      style={{ position: "absolute", top: "7px" }}
+      style={{ position: "absolute", top: "10px" }}
```
haideralsh, line 37: These values don't seem to be based on our 8px grid system, is that ok?

src/BrandedNavBar/renderSubMenuItems.tsx
```diff
@@ -67,33 +19,34 @@ const renderSubMenuLink = (subMenuItem, onItemClick) => (
...
       {subMenuItem.name}
-    </SubMenuLink>
+    </DropdownLink>
   </li>
 );
 
 const renderCustom = (subMenuItem, onItemClick) => (
-  <ApplySubMenuLinkStyles
+  <li
+    style={{ whiteSpace: "nowrap" }}
```
haideralsh: I think we should extract this to a styled component. Inline styles can make this very hard to override.

src/BrandedNavBar/renderSubMenuItems.tsx
```diff
@@ -67,33 +19,34 @@ const renderSubMenuLink = (subMenuItem, onItemClick) => (
...
+    {subMenuItem.render(onItemClick)}
+  </li>
 );
 
 const renderText = (subMenuItem) => (
-  <SubMenuText key={subMenuItem.key ?? subMenuItem.name}>
-    {subMenuItem.name}
-  </SubMenuText>
+  <li
+    style={{ whiteSpace: "nowrap" }}
```
haideralsh: I think we should extract this to a styled component. Inline styles can make this very hard to override.

src/DropdownMenu/DropdownText.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const DropdownText: React.FC<TextProps> = styled(Text)(
+  ({ theme }: TextProps): CSSObject => ({
+    color: theme.colors.darkGrey,
+    display: "block",
+    width: "100%",
+    border: "none",
+    textAlign: "left",
+    backgroundColor: "transparent",
+    transition: ".2s",
```
haideralsh, line 16: It's a good idea to specify the transition properties here, i.e what we want to animate. If none are specified it's `all` by default.

we can define it using `transition: <property> <duration>` example: `transition: background 200ms;`

src/BrandedNavBar/SubMenuTrigger.tsx
sskirby: @haideralsh unfortunately the size of the chevron Icon is not the size of the box that it's in, so we had to do custom positioning to put it at the centre of the line height.

src/DropdownMenu/DropdownText.tsx
sskirby: @haideralsh I did this by looking at what was in the `DropdownButton`, `DropdownLink`, and `DropdownItem` and made it match those. I think if I change this, they should all change. Also, I'm unsure exactly what transitions are being animated.

haideralsh: Ok I see. we can leave it the way it is.

haideralsh (approved):

src/BrandedNavBar/SubMenuTrigger.tsx
sskirby: Ok, I'm gonna leave it be then.

PR #1021, opened 2022-06-08
chore: fix duplicate story titles

sskirby:
## Description

Many storybooks under the Table component had duplicate names, causing Chromatic to not render them.

## Changes include

## Feature checklist

### DIFF
--- a/src/Table/Table.story.tsx
+++ b/src/Table/Table.story.tsx
@@ -372,7 +372,3 @@ export const WithOnHoverActions = () => {
     />
   );
 };
-
-WithOnHoverActions.story = {
-  name: "with hover actions",
-};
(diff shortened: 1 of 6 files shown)


src/Table/Table.story.tsx
```diff
@@ -374,5 +374,5 @@ export const WithOnHoverActions = () => {
 };
 
 WithOnHoverActions.story = {
-  name: "with hover actions",
+  name: "with on hover actions",
```
haideralsh: I think this could be removed all together. By default storybook takes the name of the component `WithOnHoverActions` and transforms it to a sentence case: `With on hover actions`. So we don't have to manually do this for this case.

haideralsh (approved):

src/Table/Table.story.tsx
sskirby: You are correct!

PR #1026, opened 2022-06-21
More flexible BrandedNavBar custom rendering

sskirby:
## Description

1. Adds a `trigger` prop to the BrandedNavBar which allows the trigger for menus to be customized.
2. Pass `size` into BrandedNavBar custom `trigger` and `render` functions so that they know if they are being rendered in a `medium` (regular) or `small` (hamburger) format.
3. Pass `layer` into BrandedNavBar custom `trigger` and `render` functions so that they know how deep in the menu they are being rendered.
4. Adds a convenience method to make it easy to add styled-system props to components.

## Changes include

## Feature checklist

### DIFF
--- a/src/BrandedNavBar/MenuTrigger.tsx
+++ b/src/BrandedNavBar/MenuTrigger.tsx
@@ -60,7 +60,7 @@ const MenuTrigger = ({
       trigger={() => {
         const defaultRender = () => <MenuTriggerButton {...triggerProps} />;
         return trigger
-          ? trigger({ mode: "desktop", defaultRender, layer })
+          ? trigger({ size: "medium", defaultRender, layer })
           : defaultRender();
       }}
     >
--- a/src/BrandedNavBar/MenuTrigger.tsx
+++ b/src/BrandedNavBar/MenuTrigger.tsx
@@ -1,109 +1,21 @@
 import React from "react";
-import styled, { CSSObject } from "styled-components";
-import { themeGet } from "@styled-system/theme-get";
-import { Icon } from "../Icon";
-import { DefaultNDSThemeType } from "../theme.type";
 import NavBarDropdownMenu from "./NavBarDropdownMenu";
 import SubMenuTrigger from "./SubMenuTrigger";
 import renderSubMenuItems from "./renderSubMenuItems";
+import MenuTriggerButton from "./MenuTriggerButton";
 
-export type MenuTriggerProps = {
+export type TriggerFunctionProps = {
   name?: string;
   "aria-label"?: string;
+  color?: string;
+  hoverColor?: string;
+  hoverBackground?: string;
+};
+
+export type MenuTriggerProps = TriggerFunctionProps & {
   menuData?: any[];
-  color?: string;
-  hoverColor?: string;
-  hoverBackground?: string;
+  trigger?: (props: TriggerFunctionProps) => JSX.Element;
 };
-
-type StyledButtonProps = React.ComponentPropsWithRef<"button"> & {
-  color?: string;
-  hoverColor?: string;
-  hoverBackground?: string;
-  theme?: DefaultNDSThemeType;
-};
-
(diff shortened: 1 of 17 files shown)


src/index.ts
```diff
@@ -82,3 +82,5 @@ export { ApplicationFrame, Page, Sidebar } from "./Layout";
 export { useWindowDimensions } from "./utils";
 export { Divider } from "./Divider";
 export { SortingTable } from "./SortingTable";
+export { handleStyledProps } from "./StyledProps";
```
haideralsh: Is the reason we export this is to make it possible for someone to create their own styled-component with `styled-system` attributes outside of NDS? 

If so, this might not work if the project does not include `styled-system` as a dependency no?

haideralsh: Currently the only peer dependencies we require are `react`, `react-dom` and `styled-components`.

src/BrandedNavBar/MenuTrigger.tsx
```diff
@@ -1,117 +1,30 @@
...
-);
-
-type MenuTriggerButtonProps = React.ComponentPropsWithRef<"button"> & {
-  name?: React.ReactNode;
-  color?: string;
-  hoverColor?: string;
-  hoverBackground?: string;
+export type MenuTriggerProps = TriggerFunctionProps & {
+  menuData?: any[];
+  trigger?: (props: TriggerFunctionProps) => JSX.Element;
```
haideralsh: I think it's better to use `React.ReactNode` here instead of `JSX.Element`, the latter is very limiting.

haideralsh: Also, I think `props` should be optional here since on line 60 we're not providing any arguments to both functions.

src/index.ts
sskirby: I wanted to make new styled components in Eco that extended NDS types. In order for those to work with the styled-system props (and be able to override css properties set via styled-components) I had to define my own `handleStyledProps` function in Eco (which was not as good because it didn't have the NDS extensions). So I was thinking why not have NDS export this as a utility?

But you're right, I forgot to consider the peer dependency thing. I suspect that it will be fine even if the project does not include `styled-system` because I am currently able to use `styled-system` props on NDS components in Eco and Eco didn't have a direct dependency on `styled-system` until recently.

src/BrandedNavBar/MenuTrigger.tsx
sskirby: Ok!

src/index.ts
haideralsh: I will update the documentation to list `styled-system` as an optional peer dependency if someone is to use these utility functions.

src/StyledProps/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  overflow,
+  OverflowProps,
+} from "styled-system";
+import { textOverflow, TextOverflowProps } from "./textOverflow";
+import { cursor, CursorProps } from "./cursor";
+import { transform, TransformProps } from "./transform";
+import { transition, TransitionProps } from "./transition";
+import { visibility, VisibilityProps } from "./visibility";
+
+export const handleStyledProps = compose(
```
haideralsh: Nit pick: why are we calling this `handleStyledProps`. When I see `handle...` in a React context my mind thinks event handler right away. Should this be called something else like `getStyledProps`, `withStyledProps`, `addStyledProps` or anything else you think of?

src/BrandedNavBar/TriggerFunctionProps.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+
+export type TriggerFunctionProps = {
+  mode: "mobile" | "desktop";
```
sskirby: @sskirby change these to `small` and `medium`

haideralsh (approved):

PR #1044, opened 2022-07-08
Make branded nav bar hamburger menu more extensible

sskirby:
## Description

These customization points are not available via the public API of the BrandedNavBar. They are only to be used by Eco's new NavBar.

1. Allow the BrandedNavBar left-hand logo to be customized.
2. Allow the BrandedNavBar hamburger menu trigger to be customized.
3. Add types to the `withMenuState` function.

## Changes include

## Feature checklist

### DIFF
--- a/src/BrandedNavBar/MobileMenu.tsx
+++ b/src/BrandedNavBar/MobileMenu.tsx
@@ -4,7 +4,6 @@ import { display } from "styled-system";
 import { Text, Heading3 } from "../Type";
 import { Flex } from "../Flex";
 import { BrandingText } from "../Branding";
-import { DefaultNDSThemeType } from "../theme.type";
 import { DropdownLink, DropdownText } from "../DropdownMenu";
 import { Link } from "../Link";
 import { LinkProps } from "../Link/Link";
@@ -209,10 +208,17 @@ type BaseMobileMenuProps = {
   subtext?: string;
   closeMenu?: Function;
   themeColorObject?: ThemeColorObject;
-  logoSrc?: string;
+  showNulogyLogo?: boolean;
 };
 
-const BaseMobileMenu = ({ menuData, closeMenu, subtext, themeColorObject, logoSrc, ...props }: BaseMobileMenuProps) => (
+const BaseMobileMenu: React.FC<BaseMobileMenuProps> = ({
+  menuData,
+  closeMenu,
+  subtext,
+  themeColorObject,
+  showNulogyLogo,
+  ...props
+}) => (
   <Nav backgroundColor={themeColorObject && themeColorObject.background} {...props}>
     <BrandingWrap>
       <BrandingText logoColor={themeColorObject && themeColorObject.logoColor} />
@@ -221,7 +227,7 @@ const BaseMobileMenu = ({ menuData, closeMenu, subtext, themeColorObject, logoSr
       {menuData.primaryMenu && renderTopLayerMenuItems(menuData.primaryMenu, closeMenu, themeColorObject)}
       {menuData.secondaryMenu && renderTopLayerMenuItems(menuData.secondaryMenu, closeMenu, themeColorObject)}
     </Menu>
-    {logoSrc && (
+    {showNulogyLogo && (
       <Flex textAlign="center" borderTop={borderStyle} height="40px" alignItems="center" justifyContent="center">
         <NulogyLogo />
         {subtext && (
--- a/src/BrandedNavBar/NavBar.tsx
+++ b/src/BrandedNavBar/NavBar.tsx
@@ -8,7 +8,7 @@ import { Box } from "../Box";
 import DesktopMenu from "./DesktopMenu";
 import { NulogyLogoContainer } from "./NulogyLogoContainer";
 import EnvironmentBanner from "./EnvironmentBanner";
-import BrandLogoContainer, { BrandLogoContainerProps } from "./BrandLogoContainer";
+import BrandLogoContainer from "./BrandLogoContainer";
 import SmallNavBar from "./SmallNavBar";
 import NavBarBackground from "./NavBarBackground";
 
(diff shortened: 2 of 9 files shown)


src/BrandedNavBar/MobileMenu.tsx
```diff
@@ -209,10 +208,17 @@ type BaseMobileMenuProps = {
   subtext?: string;
   closeMenu?: Function;
   themeColorObject?: ThemeColorObject;
-  logoSrc?: string;
+  showNulogyLogo?: boolean;
```
haideralsh, line 211: Are you planning to release this as a major version? since we're removing an existing prop.

src/BrandedNavBar/NavBar.tsx
```diff
@@ -23,35 +23,22 @@ const themeColorObject = {
   logoColor: "blue",
 };
 
-type MediumNavBarProps = BrandLogoContainerProps & {
-  subtext?: string;
+type MediumNavBarProps = {
   menuData?: any;
   environment?: "development" | "training";
+  logo: React.ReactElement;
```
haideralsh: I think we want to use `React.ReactNode`

src/BrandedNavBar/MobileMenu.tsx
```diff
@@ -209,10 +208,17 @@ type BaseMobileMenuProps = {
...
 
-const BaseMobileMenu = ({ menuData, closeMenu, subtext, themeColorObject, logoSrc, ...props }: BaseMobileMenuProps) => (
+const BaseMobileMenu = ({
+  menuData,
+  closeMenu,
+  subtext,
+  themeColorObject,
+  showNulogyLogo,
+  ...props
+}: BaseMobileMenuProps) => (
```
haideralsh: Can we use React.FC here to provide the children as a prop?

```suggestion
const BaseMobileMenu: React.FC<BaseMobileMenuProps> = ({
  menuData,
  closeMenu,
  subtext,
  themeColorObject,
  showNulogyLogo,
  ...props
}) => (
```

src/BrandedNavBar/NavBar.tsx
```diff
@@ -23,35 +23,22 @@ const themeColorObject = {
...
-  menuData,
-  subtext,
-  environment,
-  logoSrc,
-  brandingLinkHref = "/",
-  brandingLinkTo,
-  brandingLinkComponent,
-  ...props
-}: MediumNavBarProps) => {
+const MediumNavBar = ({ menuData, environment, logo, showNulogyLogo, subtext, ...props }: MediumNavBarProps) => {
```
haideralsh:
```suggestion
const MediumNavBar: React.FC<MediumNavBarProps> = ({ menuData, environment, logo, showNulogyLogo, subtext, ...props }) => {
```

src/BrandedNavBar/NavBar.tsx
```diff
@@ -23,35 +23,22 @@ const themeColorObject = {
   logoColor: "blue",
 };
 
-type MediumNavBarProps = BrandLogoContainerProps & {
-  subtext?: string;
+type MediumNavBarProps = {
   menuData?: any;
   environment?: "development" | "training";
+  logo: React.ReactElement;
+  showNulogyLogo?: boolean;
```
haideralsh, line 30: To avoid ending up with an impossible state where someone would pass `showNulogyLogo={true}` and `logo={<Logo />` we could use a union type here:

```
MediumNavBarLogo = { logo: React.ReactElement } | { showNulogyLogo: true; }

type MediumNavBarProps = {
  menuData?: any;
  environment?: "development" | "training";
  subtext?: string;
} & MediumNavBarLogo
```

src/BrandedNavBar/NavBar.tsx
```diff
@@ -83,19 +70,40 @@ const MediumNavBar = ({
...
 
-const SelectNavBarBasedOnWidth = ({ width, defaultOpen, breakpointUpper, ...props }: any) => {
+const SelectNavBarBasedOnWidth = ({
+  width,
+  defaultOpen,
+  breakpointUpper,
+  brandingLinkHref = "/",
+  brandingLinkTo,
+  brandingLinkComponent,
+  logoSrc,
```
haideralsh, line 87: should we change the `logoSrc` to match what the components are using `showNulogyLogo`

src/BrandedNavBar/SmallNavBar.tsx
```diff
@@ -69,32 +68,39 @@ const MenuIcon = ({ isOpen }) => {
   return <Icon icon={icon} title={title} />;
 };
 
-type SmallNavBarNoStateProps = BrandLogoContainerProps & {
-  menuState?: any;
+export type RenderMenuButtonProps = {
+  themeColorObject: any;
+  onClick: () => null;
```
haideralsh: would void be a more accurate type here `onClick: () => void`

src/BrandedNavBar/MobileMenu.tsx
sskirby: @haideralsh the public API (https://github.com/nulogy/design-system/blob/c99a2cab1e2ccf8c377887eec06238702b6e0974/src/BrandedNavBar/NavBar.tsx#L112) has not changed. This is an internal refactoring to allow Eco more control over what the logo in the upper-left of the NavBar looks like.

src/BrandedNavBar/NavBar.tsx
sskirby: @haideralsh is that because `React.ReactNode` allows a list of `React.ReactElement`s?

src/BrandedNavBar/NavBar.tsx
sskirby: @haideralsh I don't think it's an impossible state. It's conceivable that a client could provide a custom logo that's not a nulogy logo and set `showNulogyLogo` to `true`.

src/BrandedNavBar/SmallNavBar.tsx
sskirby: @haideralsh yes!

src/BrandedNavBar/NavBar.tsx
sskirby: @haideralsh Currently, `logoSrc` when is specified, we show the `logoSrc` logo in the upper-left and a Nulogy logo in the upper-right. When `logoSrc` isn't specified, we show a Nulogy logo only in the upper-left. I didn't want to change the existing behaviour of the `BrandedNavBar`, so I didn't expose the `showNulogyLogo` prop (which actually only controls if the Nulogy logo in the upper-right shows, so maybe it should be called something else?)

src/BrandedNavBar/MobileMenu.tsx
sskirby: @haideralsh I think that's a fine change, although this code was not written by me, just moved around by me.

src/BrandedNavBar/NavBar.tsx
sskirby: @haideralsh I did it, but now it takes more vertical space. Ugly :(

src/BrandedNavBar/MobileMenu.tsx
sskirby: @haideralsh technically, you are correct that since the `MobileMenu` is exported, it is part of the public API. However, only Eco is using this widget. I'll make it a breaking change if you want though. Let me know!

src/BrandedNavBar/NavBar.tsx
haideralsh: Ok I see, so we would show both logos in this case?

haideralsh (approved):

src/BrandedNavBar/NavBar.tsx
sskirby: Yes!

PR #1049, opened 2022-07-20
feat: pass along custom props to the Select component

haideralsh:
## Description
The `<Select />` currently swallows unknown props and doesn't drill them down to the `WindowedSelect` we are using underneath it. We sometimes need to pass custom props to the custom components like the `Option` and use them via `props.selectProps`. 

## Changes include

## Feature checklist

### DIFF
--- a/src/Select/Select.tsx
+++ b/src/Select/Select.tsx
@@ -132,7 +132,6 @@ const ReactSelect = forwardRef(
       required,
       requirementText,
       helpText,
-      noOptionsMessage,
       disabled,
       errorMessage,
       errorList,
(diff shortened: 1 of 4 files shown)


src/Select/Select.tsx
```diff
@@ -221,6 +221,7 @@ const ReactSelect = forwardRef(
             aria-label={ariaLabel}
             filterOption={filterOption}
             closeMenuOnSelect={closeMenuOnSelect}
+            {...props}
```
arturopie, line 200: Should we spread first, and then apply the other props? How do we want to handle prop collisions? do we want `props` to take precedence?

haideralsh: I think it makes sense for the props to take precedence, it is an escape hatch that allows you to override the default behaviour, similar to how we spread the `components` after setting the default values. What do you think?

I pushed another commit to remove the redundant props now that we're spreading all incoming props.

arturopie: Make sense!

arturopie (approved):

PR #1057, opened 2022-10-14
chore: update operating system version of release action

allank-nulogy:
## Description

This changes the release github action to use a newer operating system version that will not be deprecated soon.

## Changes include

- [X] chore: contains no changes affecting the library, such as documentation or test updates

## Feature checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
   release:
     name: Release
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
```
allank-nulogy, line 9: Also considered using `ubuntu-latest`. I assumed an explicit operating system version was chosen for safety, and proceeded according to that assumption.

PR #1060, opened 2022-10-21
fix: align checkbox check with first line of multiline labels

sskirby:
## Description

For multiline checkbox labels, the check should be aligned with the first line of text, not the entire label body.

## Changes include

## Feature checklist

### DIFF
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -6,6 +6,7 @@ import { Text } from "../Type";
 import { ClickInputLabel } from "../utils";
 import { getSubset, omitSubset } from "../utils/subset";
 import { DefaultNDSThemeType } from "../theme.type";
+import { addStyledProps } from "../StyledProps";
 
 type CheckboxProps = React.ComponentPropsWithRef<"input"> & {
   labelText?: string;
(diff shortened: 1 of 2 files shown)


src/Checkbox/Checkbox.tsx
```diff
@@ -79,7 +80,6 @@ const VisualCheckbox: React.FunctionComponent<any> = styled.div(
     borderRadius: theme.radii.small,
     border: "solid 1px",
     position: "relative",
-    alignSelf: "center",
```
haideralsh, line 82: Would changing this to `baseline` serve the same purpose as adding the `marginTop` below at line 136?

haideralsh (approved):

src/Checkbox/Checkbox.tsx
sskirby: @haideralsh not quite. Using `baseline` does align the bottom of the checkbox with the baseline of the first line of text, but the designers wanted the bottom of the checkbox slightly lower than the baseline of the first line of text. If you know a better way to do this I'm all ears though!

sskirby: Here's a link to the figma: https://www.figma.com/file/uLRNnoLgFamkSsy9zhAcPm/Capacity-Management?node-id=4283%3A157379

haideralsh: Thanks for the explanations @sskirby. I see why you went with this approach now.

haideralsh (approved):

PR #1062, opened 2022-10-24
Go 5242 nds spike copy paste a list of values on a multi select

petror-nulogy:
## Description

**Describe the change you're making, the motivations behind it, and any thing else a reviewer may need to know to approve this PR.**

## Changes include

## Feature checklist

### DIFF
--- a/cypress/integration/components/Select.spec.js
+++ b/cypress/integration/components/Select.spec.js
@@ -93,16 +93,16 @@ describe("Select", () => {
     it("copy and paste CSV labels to select", () => {
       const csvValues = "PCN2, PCN4";
 
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getMultiselect().contains("PCN2");
       getMultiselect().contains("PCN4");
     });
 
-    it("ignoring duplicated labes", () => {
+    it("ignoring duplicated labels", () => {
       const csvValues = "PCN2, PCN2, PCN1";
 
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getSelectedItems().should("have.length", 2);
 
@@ -113,7 +113,7 @@ describe("Select", () => {
     it("ignoring duplicated values that does not exist in the options list", () => {
       const csvValues = "A002, A002, B001, PCN2";
 
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getSelectedItems().should("have.length", 3);
 
@@ -125,12 +125,12 @@ describe("Select", () => {
     it("ignoring values that are already selected", () => {
       const csvValues = "PCN2, PCN1";
 
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getSelectedItems().should("have.length", 2);
 
-      const csvValues = "PCN2, PCN4";
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      const anotherCsvValues = "PCN2, PCN4";
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: anotherCsvValues });
 
       getSelectedItems().should("have.length", 3);
 
@@ -142,7 +142,7 @@ describe("Select", () => {
     it("values that are not in options list are invalid on view", () => {
       const csvValues = "PCN1, PCN5";
 
-      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getMultiselect().contains("PCN1").should("not.have.class", "invalid");
       getMultiselect().contains("PCN5").should("have.class", "invalid");
--- a/cypress/integration/components/Select.spec.js
+++ b/cypress/integration/components/Select.spec.js
@@ -93,7 +93,7 @@ describe("Select", () => {
     it("copy and paste CSV labels to select", () => {
       const csvValues = "PCN2, PCN4";
 
-      paste({ destinationSelector: selectTextInputSelector, pasteType: "text/plain", pastePayload: csvValues });
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getMultiselect().contains("PCN2");
       getMultiselect().contains("PCN4");
@@ -102,7 +102,7 @@ describe("Select", () => {
     it("ignoring duplicated labes", () => {
       const csvValues = "PCN2, PCN2, PCN1";
 
-      paste({ destinationSelector: selectTextInputSelector, pasteType: "text/plain", pastePayload: csvValues });
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
       getSelectedItems().should("have.length", 2);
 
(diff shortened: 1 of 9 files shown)


cypress/integration/components/Select.spec.js
```diff
@@ -163,3 +233,15 @@ describe("Select", () => {
   });
   describe("Knobs", () => {});
 });
+
+function paste({ destinationSelector, pastePayload, pasteType = "text/plain" }) {
```
haideralsh: Can we extract this to a spec helper file so we can use it in another context?

cypress/integration/components/Select.spec.js
```diff
@@ -163,3 +233,15 @@ describe("Select", () => {
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event
+  cy.get(destinationSelector).then(($destination) => {
```
haideralsh: Is there a reason we're prefixing this with a `$`?
```suggestion
  cy.get(destinationSelector).then((destination) => {
```

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,76 @@ describe("Select", () => {
...
+    const getSelectTextInput = () => cy.get(selectTextInputSelector);
+
+    beforeEach(() => {
+      cy.renderFromStorybook("select--paste-csv-value-in-select");
+    });
+
+    it("copy and paste CSV labels to select", () => {
+      const csvValues = "PCN2, PCN4";
+
+      paste({ destinationSelector: selectTextInputSelector, pasteType: "text/plain", pastePayload: csvValues });
```
haideralsh: The default `pasteType` is `text/plain`. Do we need to repeat it here?

petror-nulogy: Yes, we can remove it.

cypress/integration/components/Select.spec.js
petror-nulogy: I took this helper from cypress thread. There is only reason to have `$` just to define that this variable contains element, similar to jQuery style. 
But I will remove, because we do not have such style variables in the project.

src/Select/Select.tsx
```diff
@@ -126,6 +141,7 @@ const extractValue = (options, isMulti) => {
   }
 };
 
+// petror@nulogy.dev: TODO: update package "react-windowed-select" to latest version
```
haideralsh: Did you want to do this in this branch?

petror-nulogy: @haideralsh , I am not sure, maybe, it makes sense to move to another task. Just wanted to put some reminder. Because we have very old version, event without TS support.
The latest one has TS support, but requires some updates from our side too.

petror-nulogy: @haideralsh, It is not required for this task, but would be good to make this update later. I can create additional task on the board.

haideralsh: Ok in that case let's remove the comment and add this as a Jira task, since if left in the code we will not remember it.

haideralsh:
@petror-nulogy I can pair with you and go over these in more detail, but checking the stories I can see that we introduced a side-effect that causes the select to lose focus on selecting or deleting a multi-value. 

It also looks like typing and pressing enter does not have the same behaviour as pasting. We will take a look at this and also the styles for non-existent items with design.

cypress/support/commands.js
```diff
@@ -57,3 +57,15 @@ Cypress.Commands.add("isNotInViewport", (element) => {
     expect(rect.top).to.be.lessThan(0 - rect.height);
   });
 });
+
+Cypress.Commands.add("triggerPasteEvent", ({ destinationSelector, pastePayload, pasteType = "text/plain" }) {
```
haideralsh: I like adding this as a command 👍 Do you think we should change its name though to just `paste` to make the API more aligned with Cypress?

src/Select/Select.story.tsx
```diff
@@ -43,8 +43,17 @@ const wrappingOptions = [
...
-const SelectWithManyOptions = ({ multiselect, labelText }: SelectProps) => {
+const PCNList = [
+  { value: "2", label: "PCN2" },
+  { value: "4", label: "PCN4" },
+  { value: "1", label: "PCN1" },
+  { value: "9", label: "PCN9" },
+];
+
+const SelectWithManyOptions = ({ multiselect, labelText, ...props }: SelectProps) => {
+  const selectRef = useRef(null);
```
haideralsh: Do we need this ref?

src/Select/Select.story.tsx
```diff
@@ -571,3 +571,30 @@ export const WithCustomProps = () => {
...
+  };
+
+  return (
+    <Select
+      defaultValue={[partnerCompanyName[0].value, partnerCompanyName[2].value]}
+      noOptionsMessage={() => "No options"}
+      placeholder="Please select inventory status"
+      options={PCNList}
+      labelText="Select PCN"
+      className="Select"
```
haideralsh: Do we need this className?

cypress/support/commands.js
petror-nulogy: Agree.

src/Select/Select.story.tsx
petror-nulogy: className is not required, but added because other stories have it.

src/Select/Select.story.tsx
petror-nulogy: This ref is not required. Was added in the development process. Will remove it.

petror-nulogy:
@haideralsh I will check this point `I can see that we introduced a side-effect that causes the select to lose focus on selecting or deleting a multi-value.`

About this point:
`It also looks like typing and pressing enter does not have the same behaviour as pasting. We will take a look at this and also the styles for non-existent items with design.`
We do not have such requirements for current implementation. Only pasting should be handled.

petror-nulogy:
@haideralsh I pushed fix for this point: 
```
I can pair with you and go over these in more detail, but checking the stories I can see that we introduced a side-effect that causes the select to lose focus on selecting or deleting a multi-value.
```
Reason of this issue was re-render of SelectInput on every change.

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,76 @@ describe("Select", () => {
...
+    it("copy and paste CSV labels to select", () => {
+      const csvValues = "PCN2, PCN4";
+
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+
+      getMultiselect().contains("PCN2");
+      getMultiselect().contains("PCN4");
+    });
+
+    it("ignoring duplicated labes", () => {
```
haideralsh: Typo: `labes` -> `lables`

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,76 @@ describe("Select", () => {
...
+    const getSelectTextInput = () => cy.get(selectTextInputSelector);
+
+    beforeEach(() => {
+      cy.renderFromStorybook("select--paste-csv-value-in-select");
+    });
+
+    it("copy and paste CSV labels to select", () => {
+      const csvValues = "PCN2, PCN4";
+
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
```
haideralsh: Do we need to update this following renaming the command?

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,76 @@ describe("Select", () => {
...
+
+      getMultiselect().contains("PCN2");
+      getMultiselect().contains("PCN4");
+    });
+
+    it("ignoring duplicated labes", () => {
+      const csvValues = "PCN2, PCN2, PCN1";
+
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
```
haideralsh: Same here (do we need to update this following renaming the command?)

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,76 @@ describe("Select", () => {
...
+    });
+
+    it("ignoring values that are already selected", () => {
+      const csvValues = "PCN2, PCN1";
+
+      cy.triggerPasteEvent({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+
+      getSelectedItems().should("have.length", 2);
+
+      const csvValues = "PCN2, PCN4";
```
haideralsh: Should we turn this into a let since we're already defining it on line 126?

PR #1066, opened 2022-11-28
Go 5318 copy paste a list of values on a multi select

petror-nulogy:
## Description
See GO-5318 for details

## Changes include

## Feature checklist

### DIFF
--- a/cypress/integration/components/Select.spec.js
+++ b/cypress/integration/components/Select.spec.js
@@ -79,7 +79,7 @@ describe("Select", () => {
     });
   });
 
-  describe("Multiselect with copy/paste CSV functionality", () => {
+  describe("Multiselect: paste CSV sting", () => {
     const getCopyCsvLabesButton = () => cy.get("[data-testid='csv-labels']");
     const getCopyCsvValuesButton = () => cy.get("[data-testid='csv-values']");
     const getCopyCsvValuesWithDuplicateButton = () => cy.get("[data-testid='csv-labels-with-duplicates']");
@@ -110,16 +110,15 @@ describe("Select", () => {
       getMultiselect().contains("PCN2");
     });
 
-    it("ignoring duplicated values that does not exist in the options list", () => {
+    it("paste values that does not exist in the options list in input as a text", () => {
       const csvValues = "A002, A002, B001, PCN2";
 
       cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
 
-      getSelectedItems().should("have.length", 3);
+      getSelectedItems().should("have.length", 1);
 
-      getMultiselect().contains("A002");
-      getMultiselect().contains("B001");
       getMultiselect().contains("PCN2");
+      getSelectTextInput().should("have.value", "A002, B001");
     });
 
     it("ignoring values that are already selected", () => {
@@ -137,15 +136,6 @@ describe("Select", () => {
       getMultiselect().contains("PCN1");
       getMultiselect().contains("PCN2");
       getMultiselect().contains("PCN4");
-    });
-
-    it("values that are not in options list are invalid on view", () => {
-      const csvValues = "PCN1, PCN5";
-
-      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
-
-      getMultiselect().contains("PCN1").should("not.have.class", "invalid");
-      getMultiselect().contains("PCN5").should("have.class", "invalid");
     });
   });
 
--- a/cypress/integration/components/Select.spec.js
+++ b/cypress/integration/components/Select.spec.js
@@ -93,16 +93,16 @@ describe("Select", () => {
     it("copy and paste CSV labels to select", () => {
       const csvValues = "PCN2, PCN4";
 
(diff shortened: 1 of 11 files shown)


cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,66 @@ describe("Select", () => {
     });
   });
 
+  describe("Multiselect: paste CSV sting", () => {
```
haideralsh: Typo "sting" -> "string"

src/Select/Select.story.tsx
```diff
@@ -571,3 +570,29 @@ export const WithCustomProps = () => {
 UsingRefToControlFocus.story = {
   name: "using ref to control focus",
 };
+
+export const PasteCsvValueInSelect = (props) => {
```
haideralsh, line 574: Does this story provide anything over the standard multi-select one? 

If we want to keep it, should we add a sentence to explain to the person viewing it what they need to do? For instance, `paste "..." in the select input to select items`.

src/Select/Select.tsx
```diff
@@ -208,9 +246,62 @@ const ReactSelect = forwardRef(
...
+  }
+
+  if (options == null) {
+    return options;
+  } else {
+    return options.value;
+  }
+};
+
+const extractValuesFromCsvString = (csv: string) => {
```
haideralsh, line 288: Since this is a pure function can we add some tests for it? Also, optionally can we move it to a separate file e.g: `selectUtils.ts`, `selectHelpers.ts`

src/Select/Select.tsx
```diff
@@ -208,9 +246,62 @@ const ReactSelect = forwardRef(
...
+
+const getReactSelectValue = (options, input) => {
+  if (Array.isArray(input)) {
+    return input.map((i) => getOption(options, i));
+  }
+  return getOption(options, input);
+};
+
+const extractValue = (options, isMulti) => {
+  if (isMulti) {
```
haideralsh, line 277: Do we also need to check if options is null or undefined before accessing its properties?

src/Select/Select.tsx
```diff
@@ -208,9 +246,62 @@ const ReactSelect = forwardRef(
...
+};
+
+const getReactSelectValue = (options, input) => {
+  if (Array.isArray(input)) {
+    return input.map((i) => getOption(options, i));
+  }
+  return getOption(options, input);
+};
+
+const extractValue = (options, isMulti) => {
```
haideralsh, line 276: Should we add types to the parameters?

src/Select/Select.tsx
```diff
@@ -208,9 +246,62 @@ const ReactSelect = forwardRef(
     );
   }
 );
+
+const checkOptionsAreValid = (options) => {
+  if (options && process.env.NODE_ENV === "development") {
```
haideralsh, line 251: Why are we doing `process.env.NODE_ENV === "development"`?

src/Select/Select.tsx
haideralsh: I know all the other functions don't have tests, but can we add one for new ones to make it easier to reason about and change in the future?

src/Select/Select.tsx
```diff
@@ -157,16 +128,82 @@ const ReactSelect = forwardRef(
     const { t } = useTranslation();
     const themeContext = useContext(ThemeContext);
     const spaceProps = getSubset(props, propTypes.space);
+    const reactSelectRef = React.useRef<ReactSelectStateManager>(null);
 
     useEffect(() => {
       checkOptionsAreValid(options);
     }, [options]);
 
+    const handleChange = useCallback(
```
haideralsh: Nice refactoring to not recreate this function on every rerender 👏

src/Select/Select.tsx
haideralsh: I think we need to also handle having both spaces and no spaces around commas in the csv.

src/Select/Select.tsx
```diff
@@ -181,7 +218,8 @@ const ReactSelect = forwardRef(
             aria-invalid={error}
             defaultMenuIsOpen={initialIsOpen}
             inputId={id}
-            onChange={onChange && ((option) => onChange(extractValue(option, multiselect)))}
+            creatable
```
haideralsh: We can remove this.

cypress/integration/components/Select.spec.js
```diff
@@ -79,6 +79,66 @@ describe("Select", () => {
...
+    const getCopyCsvLabesButton = () => cy.get("[data-testid='csv-labels']");
```
haideralsh: Typo `labes` -> `lables`

src/Select/Select.tsx
```diff
@@ -157,16 +128,82 @@ const ReactSelect = forwardRef(
...
+
+            if (existingOption) {
+              return existingOption;
+            }
+
+            notExistingOptions.push(pastedOption);
+
+            return null;
+          })
+          .filter((pastedOption) => pastedOption)
```
haideralsh: Optional: We can replace this with `.filter(Boolean)`

haideralsh (approved):

PR #1067, opened 2022-12-02
fix: remove SelectInput rerender for options live update

petror-nulogy:
## Description

**Describe the change you're making, the motivations behind it, and any thing else a reviewer may need to know to approve this PR.**

## Changes include

## Feature checklist

### DIFF
--- a/cypress/integration/components/Select.spec.js
+++ b/cypress/integration/components/Select.spec.js
@@ -136,7 +136,7 @@ describe("Select", () => {
     });
   });
 
-  describe.only("Multiselect: options live update", () => {
+  describe("Multiselect: options live update", () => {
     const selectTextInputSelector = "[data-testid='select-input'] input";
     const getSelectTextInput = () => cy.get(selectTextInputSelector);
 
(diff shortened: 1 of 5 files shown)


cypress/integration/components/Select.spec.js
```diff
@@ -136,6 +136,39 @@ describe("Select", () => {
     });
   });
 
+  describe.only("Multiselect: options live update", () => {
```
haideralsh: I think we need to remove the only here.

cypress/integration/components/Select.spec.js
```diff
@@ -136,6 +136,39 @@ describe("Select", () => {
...
+
+    it("add new option on every input change and input should be focused", () => {
+      getSelectTextInput().type("test");
+      getSelectTextInput().should("have.focus");
+
+      const csvValues = "PCN2, PCN1";
+
+      cy.paste({ destinationSelector: selectTextInputSelector, pastePayload: csvValues });
+
+      cy.focused().type("{backspace}{backspace}{backspace}");
```
haideralsh: Why do we need to backspace before asserting if the input is focused?

cypress/integration/components/Select.spec.js
```diff
@@ -136,6 +136,39 @@ describe("Select", () => {
...
+      getSelectTextInput().should("have.focus");
+
+      cy.focused().type("{enter}");
+      cy.focused().type("{enter}");
+    });
+
+    it("user able to paste and select values with live update of options", () => {
```
haideralsh: This test seems very similar to the one above, where we type and assert focus. Is there anything else that it is testing?

petror-nulogy: Yes, good point.
I've rethought this case and left only one test case.

cypress/integration/components/Select.spec.js
petror-nulogy: Thank you! Good catch.

haideralsh (approved):

haideralsh:
We can merge this now, and come back to refactor the tests.

PR #1070, opened 2022-12-09
DES-444 export StatusIndicatorType from NDS

petror-nulogy:
## Description

Exports the StatusIndicatorValues to be used by applications using @nulogy/components

## Changes include

## Feature checklist

### DIFF
--- a/src/StatusIndicator/StatusIndicator.tsx
+++ b/src/StatusIndicator/StatusIndicator.tsx
@@ -3,48 +3,50 @@ import styled from "styled-components";
 import { space, typography, flexbox, SpaceProps, TypographyProps, FlexboxProps } from "styled-system";
 import { DefaultNDSThemeType } from "../theme.type";
 
-export enum StatusIndicatorType {
-  neutral = "neutral",
-  dark = "dark",
-  danger = "danger",
-  informative = "informative",
-  success = "success",
-  warning = "warning",
-  quiet = "quiet",
-}
+export const StatusIndicatorView = {
+  neutral: "neutral",
+  dark: "dark",
+  danger: "danger",
+  informative: "informative",
+  success: "success",
+  warning: "warning",
+  quiet: "quiet",
+} as const;
+
+export type StatusIndicatorType = typeof StatusIndicatorView[keyof typeof StatusIndicatorView];
 
 const StatusIndicatorColours = (theme) => ({
-  [StatusIndicatorType.neutral]: {
+  [StatusIndicatorView.neutral]: {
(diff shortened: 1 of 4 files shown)


src/StatusIndicator/StatusIndicator.tsx
```diff
@@ -45,19 +55,17 @@ const getStatusIndicatorColours = (type, theme) => {
   return StatusIndicatorColours(theme)[type];
 };
 
-type StatusIndicatorType = "neutral" | "dark" | "danger" | "informative" | "success" | "warning" | "quiet";
```
haideralsh, line 48: Is there a reason we're migrating away from union types? Enums have their use cases but they come with their own downsides: https://fettblog.eu/tidy-typescript-avoid-enums/

src/StatusIndicator/StatusIndicator.tsx
```diff
@@ -3,38 +3,48 @@ import styled from "styled-components";
 import { space, typography, flexbox, SpaceProps, TypographyProps, FlexboxProps } from "styled-system";
 import { DefaultNDSThemeType } from "../theme.type";
 
+export enum StatusIndicatorType {
```
haideralsh: If we have to use an enum here, I would rename it to remove the `Type` keyword, since it can't be used with standard Typescript [utility types](https://www.typescriptlang.org/docs/handbook/utility-types.html).

src/StatusIndicator/StatusIndicator.tsx
petror-nulogy: We can leave the `StatusIndicatorType`. Just can transform it to:
```
type StatusIndicatorType = keyof typeof StatusIndicatorType;
```

src/StatusIndicator/StatusIndicator.tsx
petror-nulogy: How do you like `StatusIndicatorView` instead?

src/StatusIndicator/StatusIndicator.tsx
petror-nulogy: I added and exported enum, because we want to use values directly from NDS in the ECO.

src/StatusIndicator/StatusIndicator.tsx
petror-nulogy: FYI, got propositions to use `as const` instead of `enum`. 
https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

haideralsh (approved):

PR #1078, opened 2023-02-15
fix: don't focus the element on mount

haideralsh:
## Description
Focus is currently automatically set to the first child of the `FocusManager` causing the page to be poorly accessible to non-mouse users in addition to a causing the page to scroll to the focused element if it is outside the viewport. 

## Changes include

## Feature checklist

### DIFF
--- a/src/utils/ts/FocusManager.tsx
+++ b/src/utils/ts/FocusManager.tsx
@@ -51,9 +51,8 @@ const FocusManager: React.FC<FocusManagerProps> = ({ children, refs = undefined,
 
     if (prevFocusedIndex.current !== focusedIndex) {
       updateFocused();
+      prevFocusedIndex.current = focusedIndex;
     }
-
-    prevFocusedIndex.current = focusedIndex;
   }, [focusedIndex, refs]);
 
   return (
--- a/src/utils/ts/FocusManager.tsx
+++ b/src/utils/ts/FocusManager.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useEffect, useState } from "react";
+import React, { ReactNode, useEffect, useRef, useState } from "react";
 
 type Reference = {
   current?: JSX.Element;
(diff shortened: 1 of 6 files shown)


nikola-nulogy (approved):

src/utils/ts/FocusManager.tsx
```diff
@@ -19,6 +19,7 @@ type FocusManagerProps = {
 
 const FocusManager: React.FC<FocusManagerProps> = ({ children, refs = undefined, defaultFocusedIndex }) => {
   const [focusedIndex, setFocusedIndex] = useState<number>(defaultFocusedIndex ?? 0);
+  const prevFocusedIndex = useRef<number>(focusedIndex);
```
danielsilva, line 22: Maybe we call this variable `focusedIndexRef` because the name `prevFocusedIndex` is already used by the `focusPrevious`and `focusNext`. Another option would be `focusedIndexPreviousValue`, `focusedIndexOld`, `focusedIndexAux`

src/utils/ts/FocusManager.tsx
```diff
@@ -48,7 +49,11 @@ const FocusManager: React.FC<FocusManagerProps> = ({ children, refs = undefined,
       refs[focusedIndex].focus();
     };
 
-    updateFocused();
+    if (prevFocusedIndex.current !== focusedIndex) {
+      updateFocused();
+    }
+
+    prevFocusedIndex.current = focusedIndex;
```
danielsilva: We could move this assignment into the `if` to improve semantics.

haideralsh: If we move the assignment inside the if we will not set the `prevFocusedIndex.current` on mount though, right?

src/utils/ts/FocusManager.tsx
haideralsh: The idea was to mimic `prevState` that we had in lifecycle hooks, but I see how this could be confused with `focusPrevious` function. I can rename it.

danielsilva: It's just a bit confusing because when you mount the component there is not previous focused. I'm ok if the name stays as it is. I guess it's not a big deal.

src/utils/ts/FocusManager.tsx
danielsilva: That would happen on line 22, I think (the assignment on mount-time).

haideralsh: Yes you're right, I forgot that we're initializing the ref when we're declaring it. I will move it to the `if`.

danielsilva (approved):

PR #1359, opened 2024-01-30
Add prop to disable datepicker

jherdman:
## Description

Does what it says on the tin.

## Changes include

## Feature checklist

### DIFF
--- a/src/DatePicker/DatePicker.story.tsx
+++ b/src/DatePicker/DatePicker.story.tsx
@@ -170,3 +170,19 @@ export const UsingRefToControlFocus = () => {
 UsingRefToControlFocus.story = {
   name: "using ref to control focus",
 };
+
+export const Disabled = () => (
+  <DatePicker
+    selected={select("selected", selectedDateExamples, selectedDateExamples[0], "selected")}
+    onChange={action("date changed")}
+    onFocus={action("date selector focused")}
+    onBlur={action("date selector blurred")}
+    onInputChange={action("input changed")}
+    inputProps={{ labelText: "Expiry Date" }}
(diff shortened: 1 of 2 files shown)


src/DatePicker/DatePicker.tsx
```diff
@@ -107,9 +112,7 @@ const DatePicker: React.FC<DatePickerProps> = forwardRef(
 
     const handleEnterKey = () => {
       if (ref) {
-        // @ts-ignore
```
jherdman, line 110: I couldn't figure out why these were here. Maybe the type fixes helped?

haideralsh:
Thanks @jherdman for making this change. From looking at the storybook, I can see that we disable the input but we don't change the visual appearance of it. I think we need to change that to be aligned with the other disabled inputs we have like [these inputs here](https://master--5f60c6c285eaad0022dce67f.chromatic.com/?path=/story/components-daterange--disabled).

src/DatePicker/DatePicker.story.tsx
```diff
@@ -170,3 +170,19 @@ export const UsingRefToControlFocus = () => {
...
+    onChange={action("date changed")}
+    onFocus={action("date selector focused")}
+    onBlur={action("date selector blurred")}
+    onInputChange={action("input changed")}
+    inputProps={{ labelText: "Expiry Date" }}
+    disabled
+  />
+);
+
+Default.story = {
```
haideralsh: I think you meant `Disabled.story = {...}` ?

jherdman: smh...

PR #1361, opened 2024-02-05
Update react windowed select

jherdman:
## Description

- Upgrade `react-windowed-select`
- Fix TypeScript issues related to the Select component

## Changes include

## Feature checklist


src/Select/Select.tsx
```diff
@@ -29,45 +30,40 @@ type ReactSelectStateManager = {
...
-  closeMenuOnSelect?: boolean;
-  "aria-label"?: string;
   size?: ComponentSize;
-  isClearable?: boolean;
-  [key: string]: any; // Allow for custom props to be passed and used inside custom components using the `selectProps` prop
-};
+  error?: boolean;
+  options: NDSOptionType[];
+  onChange?: (newValue: unknown) => void;
+  [key: string]: any;
```
haideralsh, line 59: This is required for backwards compatibility, we currently allow passing any random prop to the select component. We have a [story that covers this](https://master--5f60c6c285eaad0022dce67f.chromatic.com/?path=/story/components-select--with-custom-props).

This is also part of the react-select API. They do recommend typing it [Custom Select props](https://react-select.com/typescript#custom-select-props), but we don't know what props someone might pass so it can't be done in NDS.

We can remove it and not allow custom props, but we will be creating a breaking change and a new version of NDS. 


what are your thoughts @jherdman?

haideralsh:
@jherdman This PR should be good to be merged but we need to squash some of these commits before merging.

haideralsh (approved):

src/Select/Select.tsx
jherdman: IMHO let's just keep it as it maps to React Select and should result in few surprises.

haideralsh:
fix: Upgrade `react-windowed-select`

PR #1374, opened 2024-03-05
feat: add hint prop to the FieldLabel

jherdman:
## Description

FieldLabel was missing some props necessary for ECO

## Changes include

## Feature checklist

### DIFF
--- a/src/FieldLabel/Label.tsx
+++ b/src/FieldLabel/Label.tsx
@@ -1,14 +1,18 @@
 import styled from "styled-components";
-import { space, color } from "styled-system";
-import PropTypes from "prop-types";
+import { space, color, display, DisplayProps } from "styled-system";
+import type { SpaceProps } from "styled-system";
+import type { ComponentPropsWithRef } from "react";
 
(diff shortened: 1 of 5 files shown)


src/FieldLabel/Label.tsx
```diff
@@ -1,15 +1,14 @@
 import styled from "styled-components";
 import { space, color } from "styled-system";
-import PropTypes from "prop-types";
+import type { ColorProps, SpaceProps } from "styled-system";
+import type { ComponentPropsWithRef } from "react";
 
-const Label = styled.label(space, color, () => ({
+export type Props = ColorProps & SpaceProps & ComponentPropsWithRef<"label">;
```
haideralsh: Is there a reason we're using type here instead of interface?

jherdman: Yeah, I had mistakenly assumed (see latest commit) that ESLint would complain that the interface is empty. It's not, so it's OK with `interface`.

haideralsh (approved):

PR #1406, opened 2024-05-08
fix: allow passing metadata to the table columns

haideralsh:
## Description
Allow passing metadata to the table columns

## Changes include

## Feature checklist

### DIFF
--- a/src/Table/Table.types.ts
+++ b/src/Table/Table.types.ts
@@ -3,32 +3,26 @@ import PropTypes from "prop-types";
 
 export type RowType = unknown;
 
-export interface CellInfoType {
+export interface CellInfoType<ColumnMetadata> {
   cellData: unknown;
-  column: ColumnType;
+  column: ColumnType<ColumnMetadata>;
   row: RowType;
-}
-
-interface ColumnInfoType {
-  align?: ColumnAlignment;
-  label: string;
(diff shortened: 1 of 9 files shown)


src/Table/Table.types.ts
```diff
@@ -9,23 +9,17 @@ export interface CellInfoType {
...
-interface ColumnInfoType {
-  align?: ColumnAlignment;
-  label: string;
-  dataKey?: Key;
-  width?: string | number;
-}
-
 type ColumnAlignment = "left" | "right" | "center";
 
 export type ColumnType = {
```
jherdman: IMHO we need to offer a generic to express the metadata type lest we end up with scenarios like this:

```typescript
/**
  * assume metadata is... { actionText: "Button clicked" }
  */

const buttonRenderer = ({ label, metadata }: ColumnType) => (
  <Button onClick={action(metadata.actionText)}>{label}</Button> // Argument of type 'unknown' is not assignable to parameter of type 'string'
);
```

jherdman (changes requested):

src/Table/Table.types.ts
haideralsh: That's a good idea. I pushed a change.

jherdman (approved):

PR #1417, opened 2024-05-23
Fix Popper Timeout Tracking

jherdman:
## Description

We noticed that Popper was sometimes closing the dorpdown inaccurately.
This was pinned down to timeouts being incorrectly cleared, and thefore
sometimes running when they ought not to. Careful examination of the
code revealed that they weren't being tracked very well, and thus stomped on
or lost.

This approach leverages a ref to track the value to avoid loss during
re-renders, and to also avoid unnecessary re-renders.

## Changes include

## Feature checklist

### DIFF
--- a/src/Popper/Popper.tsx
+++ b/src/Popper/Popper.tsx
@@ -50,24 +50,29 @@ const Popper: React.FC<PopperProps> = React.forwardRef(
     popperRef
   ) => {
     let timeoutID;
+
     const [isOpen, setIsOpen] = useState(defaultOpen);
-    const conditionallyApplyDelay = (fnc, delay, skipDelay = true) => {
+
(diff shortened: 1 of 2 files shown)


src/Popper/Popper.tsx
```diff
@@ -49,42 +49,59 @@ const Popper: React.FC<PopperProps> = React.forwardRef(
...
+
+    const closePopUp = (skipDelay: boolean) => {
       setPopUpState(false, skipDelay);
     };
+
     useEffect(() => {
-      const handleKeyDown = (event) => {
-        switch (event.keyCode) {
+      const handleKeyDown = (event: KeyboardEvent) => {
+        switch (event.code) {
```
jherdman: This change was a result of typing the event. Once this was fixed the types informed me (correctly) that `keyCode` is deprecated in favour of `code`.

PR #1443, opened 2024-09-20
fix: do not use an <hr /> as a divider

haideralsh:
## Description
Replaces the underlying HTML for the <Divider /> component while keeping the same API.

## Changes include

## Feature checklist


jherdman:
I'm curious about the motivation. `<hr` is defined as a thematic break between paragraphs. Is the motivation purely semantic or otherwise?

haideralsh:
@jherdman I saw that we have many instances where we use the divider for visual spacing purposes, improving readability or grouping related fields, rather than to indicate major thematic breaks.

PR #1447, opened 2024-10-08
feat: Add support for 'disabled' prop to the DatePicker component

jchunky:


haideralsh:
Thanks for the change @jchunky. It looks like we've introduced a bug with this change in the DateRange component. You can see in this [story](https://5f60c6c285eaad0022dce67f-rgfkrkoezn.chromatic.com/?path=/story/components-daterange--disabled) that the component is no longer disabled.

And here is the [before and after](https://www.chromatic.com/test?appId=5f60c6c285eaad0022dce67f&id=670588345f42766d0f8830b2).

jchunky:
We found that we could just pass 'disabled' to the 'inputProps' to resolve our issue that was motivating this pull request.

jchunky:
We don't need 'disabled' prop on the DatePicker, since we can pass 'disabled' via 'inputProps'.

PR #1468, opened 2024-10-23
Add a touch variant

haideralsh:
## Description

BREAKING CHANGE: changes NDSProvider `size` prop to `variant`.

The `size` prop was originally used sparingly in some components like the
Button and the Icon to resize those components.

It was later extended to make all interactive components large enough to
be used on a touch screen, through changing the `size` prop directly or
by passing a `size` prop to the NDSProvider globally.

With this change, we retain the use of the `size` prop for select
components, and introduce a `variant` prop that can be passed either to
individual components or the NDSProvider globally with the value of
either `desktop` or `touch`.

## Changes include

## Feature checklist


src/Alert/Alert.story.tsx
```diff
@@ -8,9 +8,9 @@ export default {
 };
 
 const alertTypes = ["danger", "informative", "success", "warning"] as const;
-
 export const AlertTypes = () => (
   <Flex flexDirection="column" gap="x1">
+    {" "}
```
haideralsh, line 13: Remove in next version

PR #1509, opened 2025-01-17
feat: implement select all deselect all dropdown select with checkbox in nds

allank-nulogy:
## Description

Adds a select component with checkboxes for each item and the ability to select all

## Changes include

- [X] feature: a non-breaking change that adds functionality

## Feature checklist

- [X] Documentation has been updated

### DIFF
--- a/src/CheckboxSelect/CheckboxSelect.spec.tsx
+++ b/src/CheckboxSelect/CheckboxSelect.spec.tsx
@@ -12,19 +12,19 @@ describe("CheckboxSelect", () => {
   it("displays the selected option if only one is selected", () => {
     const component = new Component({ value: [2] });
 
-    expect(component.triggerInput).toHaveDisplayValue("Line 2 (1/3)");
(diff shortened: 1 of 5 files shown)


src/CheckboxSelect/CheckboxSelect.spec.tsx
```diff
@@ -0,0 +1,53 @@
+import React from "react";
```
jherdman, line 1: Should this be using Cypress instead?

src/CheckboxSelect/CheckboxSelect.tsx
```diff
@@ -0,0 +1,144 @@
...
+    });
+  } else {
+    return t("Multiple ({{selectedCount}}/{{total}})", {
+      selectedCount: value.length,
+      total: options.length,
+    });
+  }
+}
+
+export type CheckboxSelectOption = {
```
jherdman, line 41: Convert to interface?

PR #1510, opened 2025-01-21
Add a Month picker & other fixes

haideralsh:
## Description
* Remove unused package
* Reduce vertical padding in the Switcher
* Bump react-datepicker
* Add a month picker
* changelog had two 11.0.0 entries
* make the MonthPicker a standalone component
* Restructure DatePicker directory
* Improve the contribution guide
* Limit React support to version 16 to 18
* Remove `ts-nocheck` from the DateRange component

## Changes include

## Feature checklist


src/DatePickers/DatePicker.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface DatePickerProps extends BaseDatePickerProps {
+  highlightDates?: ReactDatePickerProps["highlightDates"];
+}
+
+const DEFAULT_DATE_FORMAT = "yyyy-MMM-dd";
+const DEFAULT_PLACEHOLDER = "YYYY-Mon-DD";
+
+const DatePicker = forwardRef<unknown, DatePickerProps>(({ highlightDates, ...props }, datePickerRef) => {
+  const handleDownKey = () => {
+    console.log("inside the date picker");
```
jherdman: Don't forget to remove this

src/DatePickers/components/DatePickerHeader.tsx
```diff
@@ -0,0 +1,85 @@
...
+  date,
+  increaseYear,
+  decreaseYear,
+  prevMonthButtonDisabled,
+  nextMonthButtonDisabled,
+  locale,
+}: ReactDatePickerCustomHeaderProps & { locale?: string }) {
+  const { t } = useTranslation();
+
+  console.log(t("go to previous year"));
```
jherdman: This one too

src/DatePickers/components/MonthPickerHeader.tsx
```diff
@@ -0,0 +1,40 @@
...
+  date,
+  increaseYear,
+  decreaseYear,
+  prevMonthButtonDisabled,
+  nextMonthButtonDisabled,
+  locale,
+}: ReactDatePickerCustomHeaderProps & { locale?: string }) {
+  const { t } = useTranslation();
+
+  console.log(t("go to previous year"));
```
jherdman: One more

src/DatePickers/shared/BasePicker.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    const customInputProps = {
+      ...InputFieldDefaultProps,
+      inputWidth: componentVariant === "touch" ? "280px" : "184px",
+      error: !!(errorMessage || errorList),
+      ...inputProps,
+      placeholder: (inputProps && inputProps.placeholder) || defaultPlaceholder,
+    };
+
+    const customInput = (
```
jherdman, line 93: Shouldn't this be defined outside of the component body?

jherdman:
I'm a little worried about those unknowns floating around here. Can they be narrowed at all?

src/DatePickers/shared/BasePicker.tsx
haideralsh: I thought it's valid to define elements inside of a component body so long as they need access to props/state. It's only components that shall not be defined inside components as far as I understand.

haideralsh:
> I'm a little worried about those unknowns floating around here. Can they be narrowed at all?

We can narrow it down and I did in the most recent commit. I am worried that it might create breaking changes for people since it was unknown in the past. I am trying to balance correctness with backwards compatibility.

src/DatePickers/shared/BasePicker.tsx
jherdman: I could have sworn there was a React Rule about this, but I cannot find it. I thought the challenge with inline definitions like this was that you effectively recreate the component on each re-render?

haideralsh: I think that's correct in the case of a component (function), however, I've seen many libraries defining elements (constants) inside component bodies. 

I believe the issue with components inside components is that the state will be flushed at every re-render which can be problematic. 

Edit: Actually I am not sure about this anymore. ChatGPT thinks both are ok, which I am 99% sure is not the case:
https://chatgpt.com/share/6791459f-ddd8-800a-859e-d12bad980a08

PR #1593, opened 2025-06-10
feat: replace the underlying component of the Tooltip

haideralsh:
## Description

Changes the underlying component of the Tooltip to be based on [Radix](https://www.radix-ui.com/primitives/docs/components/tooltip)

BREAKING CHANGE:

* Removes the ability to pass a ref to the Tooltip component
* Removes the hideDelay prop on the Tooltip

## Changes include

## Feature checklist


jherdman:
Well, this is one way to solve the problem 😂, but it is the right way.

haideralsh:
> Well, this is one way to solve the problem 😂, but it is the right way.

We preserve 98% of the API and get rid of 4.5 year old version of Popper! what's not to like?! 😆

PR #1595, opened 2025-06-25
Upgrade husky

jherdman:
## Description

This is a small step towards brining this project into the present.

## Changes include

## Feature checklist


.github/workflows/pull-request.yml
```diff
@@ -13,18 +13,22 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
sskirby, line 18: Is this the new default for the checkout action? IIRC this was important so the whole repo history wasn't downloaded by the action.

jherdman: Indeed! It's sort of backwards actually. A depth of zero means (I kid you not): fetch the entire history.

jherdman: Reference: https://github.com/actions/checkout

PR #72, opened 2018-10-16
Nds 348 documenting how to write component documentation

nikola-nulogy:
### What I did

- Changed IA of Designers Guide
- Added Writing Component Documentation article under Writing Component Documentation

Titles under Designers Guide may be too long. Maybe we can bounce off some ideas on how to shorten them.

PR #1722, opened 2026-03-06
feat: add IBM Plex Sans JP custom font for ja_JP locale

nikola-nulogy:
## Description

Adds IBM Plex Sans JP as the custom font for the `ja_JP` locale, following the same pattern as Simplified Chinese (Noto Sans SC). The font family token is sourced from `@nulogy/tokens` v6.2.0 (via [nds-tokens PR #72](https://github.com/nulogy/nds-tokens/pull/72)), which was bumped as part of this change. README has been updated to instruct consumers to load the font if their application supports Japanese.

## Changes include

## Feature checklist


jherdman (approved):
lgtm but it's not really clear to me if we need to add instructions for applications consuming NDS to access this font

jherdman (approved):

PR #1801, opened 2026-06-05
fix(DatePickers): localize header from context and portal the calendar to escape modals

allank-nulogy:
## Description

Two fixes to the `DatePicker`, both surfaced while localizing date pickers into a new locale.

**1. The month/year header didn't localize from context.** The custom header rendered its caption from the explicit `locale` prop, which is `undefined` when a consumer relies on the `NDSProvider` context (the common case). So the caption fell back to English (e.g. "June 2026") while the rest of the calendar — weekday headers, etc. — localized correctly. `DatePickerHeader` / `MonthDatePickerHeader` now fall back to the `LocaleContext` locale (`locale ?? useLocale().locale`), mirroring what `BasePicker` already does for the picker itself.

**2. The calendar was clipped inside modals.** The calendar popper rendered inline, so a `Modal`'s `overflow: hidden` plus its `transform` containing block clipped the popup (a `strategy: "fixed"` popper can't escape a transformed ancestor). The calendar now renders in a portal (`portalId="nds-date-picker-portal"`); the popper/calendar styles were moved off `.nds-date-picker` onto the portal node so they still apply, and the popper's `z-index` was raised from `openControl` to `aboveOverlay` so it stacks above modals.

A new Modal story (**"with date picker"**) opens a calendar inside a short modal and leaves it open, so Chromatic captures the calendar escaping the modal — both as visual proof and as a regression guard.

## Changes include

## Feature checklist

- [x] Appropriate tests have been added — new Modal "with date picker" story with a `play` that opens the calendar; existing DatePicker/MonthPicker/WeekPicker (21) and Modal (11) story play-tests pass; `tsc` and `biome` clean.
- [x] Documentation has been updated — the new Storybook story documents the date-picker-in-modal case.
- [x] Accessibility has been considered — portal keeps the calendar in the DOM and reachable; arrow-button labels unchanged; z-index raised so the open calendar isn't obscured by the modal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/DatePickers/shared/components/BasePicker.tsx
+++ b/src/DatePickers/shared/components/BasePicker.tsx
@@ -145,6 +145,7 @@ export const BasePicker = forwardRef<ReactDatePicker, BasePickerProps>(
           onEnterKeyPress={onEnterKeyPress}
           popperPlacement="bottom-start"
           popperProps={getPopperProps(disableFlipping)}
+          portalId="nds-date-picker-portal"
           disabledKeyboardNavigation={disabledKeyboardNavigation}
         />
         <InlineValidation mt="x1" errorMessage={errorMessage} errorList={errorList} />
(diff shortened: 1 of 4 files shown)


jherdman (approved):
nice catch

PR #1802, opened 2026-06-09
fix(DatePickers): keep the calendar clickable inside a modal

allank-nulogy:
## Problem

In the Modal "with date picker" story, clicking a day in the calendar closed the modal instead of selecting the date.

## Root cause

Commit ef02d8af moved the DatePicker calendar into a portal appended to `document.body` so it could escape the modal's `overflow: hidden`. But the Modal is a Radix **modal** `Dialog`, which sets `pointer-events: none` on everything outside the dialog content.

Because the portal lives outside the dialog subtree, the calendar's day cells inherited `pointer-events: none` — so clicks fell straight through the dead calendar to the `Dialog.Overlay` underneath, which Radix reads as an outside click and closes the modal.

## Fix

Set `pointer-events: auto` on the `#nds-date-picker-portal` node so the portaled calendar stays interactive inside a modal. Clicks now land on the day and never reach the overlay.

## Test

The `WithDatePicker` story's `play()` previously stopped at opening the calendar, which is why this slipped through. It now clicks a day and asserts the dialog stays open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Modal/Modal.story.tsx
+++ b/src/Modal/Modal.story.tsx
@@ -238,6 +238,12 @@ export const WithDatePicker: Story = {
     const body = within(document.body);
     await userEvent.click(body.getByLabelText("select a date"));
     await waitFor(() => expect(document.querySelector(".react-datepicker-popper")).toBeInTheDocument());
+
+    // Selecting a day must not close the modal: the calendar renders in a portal
+    // outside the modal's DOM, so the dismissable layer must ignore it.
+    const day = document.querySelector(".react-datepicker__day--015:not(.react-datepicker__day--outside-month)");
+    await userEvent.click(day as Element);
+    expect(body.getByRole("dialog")).toBeInTheDocument();
   },
 };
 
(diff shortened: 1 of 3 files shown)


src/Modal/Modal.story.tsx
```diff
@@ -238,6 +238,12 @@ export const WithDatePicker: Story = {
     const body = within(document.body);
     await userEvent.click(body.getByLabelText("select a date"));
     await waitFor(() => expect(document.querySelector(".react-datepicker-popper")).toBeInTheDocument());
+
+    // Selecting a day must not close the modal: the calendar renders in a portal
+    // outside the modal's DOM, so the dismissable layer must ignore it.
+    const day = document.querySelector(".react-datepicker__day--015:not(.react-datepicker__day--outside-month)");
```
allank-nulogy, line 244: LMK if you want to remove this. It will close the modal, which may be weird for the person looking at the story

jherdman (approved):

PR #1814, opened 2026-07-08
feat: add ScrollTimePicker and DateTimePicker components

yusaira-khan:
## Description

Adds two new components for entering dates and times, both designed to behave predictably in UTC (the original goal of CN1-5260: a picker that looks like the default Chrome time picker but operates in UTC time).

- **`ScrollTimePicker`** — a scroll-wheel time picker modeled on the native Chrome time control, with hour/minute (and optional seconds) columns. Supports only 24h display, a `utc` flag so the emitted value is interpreted/emitted in UTC rather than browser-local time, keyboard and pointer selection, and a caret-driven input. Ships with time parsing/formatting utilities.
- **`DateTimePicker`** — composes the existing `DatePicker` with `ScrollTimePicker` and merges the two fields into a single UTC-aware `Date` instant. Controlled and uncontrolled modes, `utc` and `showSeconds` flags, min/max date bounds, validation, and per-sub-field escape hatches (`dateInputProps`, `timeProps`).

Both are exported from `src/index.ts`. i18n keys for the new labels were added across all locale files. **(except for "select a date")**

### Reviewer notes
- The controlled/uncontrolled sync in `DateTimePicker` follows the loop-safe patterns already used by `DatePicker` (`selected` effect) and `DateRange` (`emitRange`) — the controlled `useEffect` is a pure external→internal sync and never calls `onChange`.
- This started as a spike; the components are complete with tests and accessibility attributes.
- The exisiting DatePicker's "select a date" label didnot have i18n keys. and none were added.

## Changes include

## Feature checklist


nikola-nulogy:
@yusaira-khan Looks good. I do have a comment related to ScrollTimePicker.

1. Can we make the scrollable dropdown opens when the focus is on the input. The same the calendar opens when focus gets on DatePicker?

Other than that looks good to me. Thank you for doing this.

src/DateTimePicker/DateTimePicker.utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  // The day to feed react-datepicker as `selected`. In UTC mode this is a *proxy* Date whose
+  // LOCAL y/m/d equal the instant's UTC y/m/d, so react-datepicker (which reads local getFullYear/
+  // getMonth/getDate) renders the UTC calendar day. In local mode it is the instant itself.
+  dateForPicker: Date | undefined;
+  // The time-of-day as an "HH:mm[:ss]" string for ScrollTimePicker.
+  time: string;
+};
+
+function isValidDate(value: Date | undefined | null): value is Date {
+  return value instanceof Date && !Number.isNaN(value.getTime());
```
jherdman, line 23: You can have a date object whose time value is NaN!?

src/DateTimePicker/DateTimePicker.utils.ts
```diff
@@ -0,0 +1,52 @@
...
+//
+// The one net-new problem this component has is merging a calendar day (from DatePicker) with a
+// time-of-day (from ScrollTimePicker) into a single instant, correctly under either local time or
+// UTC. Time extraction/formatting is delegated to the ScrollTimePicker utils — we do not duplicate
+// it here.
+//
+// Convention (shared with ScrollTimePicker): the `utc`/`showSeconds` boolean flags are never
+// positional — they are always passed via an options object so call sites read clearly.
+
+import { dateToTimeParts, formatTime, parseTime, type TimeOptions } from "../ScrollTimePicker/ScrollTimePicker.utils";
```
jherdman, line 11: IMHO if these utils are shared across components they need to be lifted up to a better location.

src/DateTimePicker/DateTimePicker.utils.ts
yusaira-khan: according to claude:
>date.getTime() returns NaN whenever the Date is an Invalid Date — i.e. its internal time value is NaN.                                                                                                                            
>                                                                                                                                                                                                                                         
>  That happens when a Date is constructed from something unparseable, for example:                                                                                                                                                       
>                                                                                                                                                                                                                                         
>  new Date("not a date").getTime()      // NaN                                                                                                                                                                                           
>  new Date("2026-13-45").getTime()      // NaN  (out-of-range parts)                                                                                                                                                                     
>  new Date(undefined).getTime()         // NaN                                                                                                                                                                                           
>  new Date(NaN).getTime()               // NaN                                                                                                                                                                                           
>  new Date("").getTime()                // NaN

src/DateTimePicker/DateTimePicker.utils.ts
yusaira-khan: DateRange does something similar with TimePicker utils

PR #1826, opened 2026-07-16
fix: stop styled-system props leaking onto the DOM via styled(component) wrappers

### LINKED ISSUE #1825: AnimatedBox leaks styled-system props to the DOM (React unrecognized-prop warnings)
## Summary

`AnimatedBox` forwards styled-system props (`marginTop`, `flexShrink`, `borderRadius`, `boxShadow`, etc.) straight to the DOM, producing React console warnings like:

```
React does not recognize the `marginTop` prop on a DOM element. If you intentionally
want it to appear in the DOM as a custom attribute, spell it as lowercase `margintop`
instead...
```

Reproduce: open any story that renders a `Toggle` (e.g. `FieldLabel > With Range Of Input Components`) and watch the console. Pre-existing on `main` (independent of the React 19 upgrade — React emits this warning identically on React 18 and 19).

## Root cause

`AnimatedBox` is a styled framer-motion component (`src/Box/Box.tsx`):

```ts
export const AnimatedBox = styled(motion.div)<AnimatedBoxProps>(addStyledProps);
```

The global prop filter in `src/theme/NDSThemeProvider.tsx` only filters when the styled target is a raw DOM string tag:

```ts
function shouldForwardProp(propName, target) {
  if (typeof target === "string") {
    return isPropValid(propName);
  }
  return true; // component targets (e.g. motion.div) → forward everything
}
```

Because `AnimatedBox`'s target is the `motion.div` *component* (not a string), every prop takes the `return true` branch. The styled-system props then pass through `motion.div` onto the real DOM `<div>`, and React warns. Plain `styled.div` components are unaffected (string target → `isPropValid` filters them).

`Toggle`'s `Switch` (`src/Toggle/ToggleButton.tsx`) renders an `AnimatedBox` with `marginTop`, `flexShrink`, `borderRadius`, and `boxShadow` — the exact props in the warnings.

## Why the naive fix doesn't work

Filtering component targets with `isPropValid` too would also strip framer-motion's own props (`animate`, `whileHover`, `variants`, `initial`, `transition`) — they're not valid DOM attributes either — breaking the animations.

## Fix options

- Give `AnimatedBox` a scoped `.withConfig({ shouldForwardProp })` that forwards valid DOM props (`isPropValid`) **plus** an allow-list of framer-motion props, and drops the rest.
- Or strip styled-system prop keys from what reaches `motion.div` (e.g. via the styled-system `props` metadata) before forwarding.

## Scope

Console-only warning — no test or build failure. Affects all `AnimatedBox` consumers (Toggle, and any other `motion`-backed styled components with the same pattern). Worth auditing other `styled(motion.*)` usages while fixing.

jherdman:
## Description

Fixes the React `does not recognize the \`X\` prop on a DOM element` console warnings (`marginTop`, `flexShrink`, `borderRadius`, `boxShadow`, `maxWidth`, `paddingLeft`, `flexDirection`, …) seen while browsing stories.

**Root cause:** NDSProvider's global `shouldForwardProp` only filters string/DOM targets (`typeof target === "string"`). Any `styled(SomeComponent)` — `styled(motion.*)`, `styled(Radix.*)`, `styled(Svg)` — hits the `return true` branch and forwards **every** prop, so styled-system props pass straight through the wrapped component onto the DOM element (Radix's `asChild` makes this worse by merging them onto the rendered child).

**Fix:** per-component `shouldForwardProp`, following the existing `excludeStyledProps` pattern (already used by `BottomSheet`):
- **AnimatedBox** — `isValidMotionProp(prop) || isPropValid(prop)` (keep framer-motion + DOM props, drop the rest). A styled-system *denylist* can't be used here: framer-motion's `transition`/`transform` props collide with styled-system prop names and would be stripped, breaking animations.
- **Tooltip `StyledContent`, `Icon`, and Radix Navigation menu components** (`NavigationMenuLink`/`NavigationMenuTrigger`, `UserMenu.Container`) — `excludeStyledProps(...)`.

A **global** fix was rejected: styled-system prop names like `size`/`color`/`width` are legitimate real props on some components (e.g. `Icon` consumes `size`), so a blanket denylist would break them.

Styled-system props are still applied as styles — `shouldForwardProp` only controls DOM forwarding — so there's no visual change.

### Validation
The full Storybook suite went from **12 → 0** of these warnings. `pnpm check` clean; 610 story + 126 unit tests pass.

Closes #1825

## Changes include

## Feature checklist


jherdman:
Also folded into this PR: **re-enabling Chromatic visual snapshots**.

While investigating I found `.storybook/preview.tsx` had `chromatic: { disableSnapshot: true }` set **globally** (since commit `3ca7cfa5`, 2026-03), so CI has been publishing Storybook but taking **zero snapshots** — no visual regression testing has actually run for ~4 months. That's how the recent DatePicker font regression and these prop leaks went unnoticed. Removed it to restore the 'every story is snapshotted on CI' behaviour.

Heads-up for reviewers:
- The first Chromatic build here will snapshot all ~598 stories and likely surface a large batch of changes to review (4 months of accumulated drift + this PR's DOM changes — though the prop-leak fix should be visually neutral).
- TurboSnap (`onlyChanged`) is currently defeated because `preview.tsx` imports the package barrel (`../src`), so builds will be full snapshots until that's addressed separately. Happy to file a follow-up for it.

jherdman:
Follow-up from my previous comment — both Chromatic improvements are now **included in this PR** (no separate issue):

- **TurboSnap fixed** — `preview.tsx` now imports `NDSProvider`/`ALL_NDS_LOCALES` from their specific modules instead of the `../src` barrel, so a change to one component no longer forces a full snapshot of all ~598 stories. `onlyChanged` can finally narrow to affected stories.
- **PRs now gate on visual changes** — dropped `exitOnceUploaded` so the Chromatic check reflects the actual diff result (red until changes are reviewed), and added `autoAcceptChanges: main` so post-merge builds re-baseline cleanly.

Caveat unchanged: this PR's *first* Chromatic build is still a full one (the changed files — Box/Icon/Tooltip/Navigation + preview.tsx — are broad), and it'll surface the ~4 months of accumulated drift for review. Subsequent PRs benefit from TurboSnap. Also worth confirming the `Chromatic` check is marked **required** in branch protection so the gate actually blocks merges.

PR #1832, opened 2026-07-22
SPAR-444: publish @nulogy/components to GitHub Packages

sskirby:
## What

Switch `@nulogy/components` publishing from public npmjs to **GitHub Packages** (`npm.pkg.github.com`), keeping the `@nulogy` scope.

Part of [SPAR-444](https://nulogy-go.atlassian.net/browse/SPAR-444). NPM allows only one registry per package scope, so `@nulogy/*` on npmjs and the private `@nulogy/spark-ui` on GitHub Packages cannot both install in PackManager. Consolidating the NDS family (`@nulogy/components`, `@nulogy/icons`, `@nulogy/tokens`) onto GitHub Packages resolves the conflict without exposing spark-ui publicly.

## Changes

- `package.json` → `publishConfig.registry = https://npm.pkg.github.com`, `access: restricted`
- `.github/workflows/release.yml`: add `packages: write`; remove `id-token: write` (npm OIDC trusted publishing is npmjs-only, inert for GitHub Packages); authenticate `@semantic-release/npm` via `NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}` (the plugin reads `NPM_TOKEN`, not `NODE_AUTH_TOKEN`)
- README: install from the `@nulogy` GitHub Packages registry

## Publishing sequencing

- **Merging this does not publish.** The commit is `ci:`-typed, so semantic-release cuts no release on merge. The first GitHub Packages release happens on the next releasable (`feat:`/`fix:`) commit to `main`; the version continues from existing git tags (stays on the current major), satisfying PackManager's current range.
- `npm whoami` is skipped for non-default registries, so verify needs only `NPM_TOKEN` present (@semantic-release/npm `verify-auth.js`).

## Related

- Companion PRs (all three must land together): `nulogy/design-system`, `nulogy/nds-icons`, `nulogy/nds-tokens`
- Follow-up (PackManager): point the `@nulogy` scope at GitHub Packages and drop the `file:` spark-ui dependency.

[SPAR-444]: https://nulogy-go.atlassian.net/browse/SPAR-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,9 +17,7 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
-      # Required for OIDC to let NPM registry know
-      # that this workflow is trusted
-      id-token: write
+      packages: write
     steps:
       - name: Checkout
         uses: actions/checkout@v7
@@ -52,3 +50,5 @@ jobs:
           branch: main
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # @semantic-release/npm authenticates via NPM_TOKEN, not NODE_AUTH_TOKEN
+          NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/mirror-to-github-packages.yml
+++ b/.github/workflows/mirror-to-github-packages.yml
@@ -1,0 +1,41 @@
+name: Mirror versions to GitHub Packages
+
+# One-off (SPAR-444): republish existing npmjs versions of this package to
+# GitHub Packages. semantic-release only publishes versions forward, so the
+# older versions consumers still pin have to be mirrored for them to install
+# from GitHub Packages.
+
+on:
+  workflow_dispatch:
+    inputs:
+      versions:
+        description: "Space-separated npmjs versions to mirror (e.g. 19.1.3 15.1.4)"
+        required: true
+
+permissions:
+  packages: write
+
+jobs:
+  mirror:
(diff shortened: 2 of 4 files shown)


jherdman:
> Yarn allows only one registry per package scope, so...

My dude... [pnpm](https://pnpm.io/).

.github/workflows/release.yml
```diff
@@ -17,9 +17,7 @@ jobs:
       contents: write
       issues: write
       pull-requests: write
-      # Required for OIDC to let NPM registry know
-      # that this workflow is trusted
-      id-token: write
```
jherdman, line 22: Are you sure this didn't break our existing publishing workflow?

jherdman (commented):
Do you actually need to make these changes? This is a public package after all.

ashik-v (approved):

.github/workflows/release.yml
sskirby: I am moving everything to Github NPM because we can't have two registries with the same name (i.e. Nulogy) and we don't want to make all of our npm packages public.

sskirby:
> > Yarn allows only one registry per package scope, so...
> 
> My dude... [pnpm](https://pnpm.io/).

I will fix that. It actually has nothing to do with the package manager, it's an NPM thing.

